PR #97, opened 2023-08-05
Add the folder structure and two files

GeorgeKerscher:
I created the folder structure Matt recommended. I added a copyright statement in the experiments  folder and a html template in the tactile-graphics folder. I expect these files will need to change, and I will email them to William and Nicole for review.

### DIFF
--- a/experiments/copyright-statement.html
+++ b/experiments/copyright-statement.html
@@ -1,0 +1,24 @@
+﻿<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" lang="en-US" xml:lang="en-US">
+	<head>
+		<meta charset="utf-8" />
+<head>
+
+<title>Copyright Information about Experiments and Samples</title>
+
+<meta name="author" content="George Kerscher" />
+
+<link rel="stylesheet" href="common/css/common.css" />
+
+</head>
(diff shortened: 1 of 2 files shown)


mattgarrish:
I think it might be better to convert the copyright page to a readme file, that way it will be displayed whenever anyone navigates to the folder in github. But I can do that after you merge if you want.

mattgarrish (approved):

GeorgeKerscher:
Hi,

 

I thought everybody would see the copyright statement when they navigated to the folder. If they would not see it, but would see the readme, then that would be better. I just thought the copyright statement would be informative.

 

Best

George

mattgarrish:
> I thought everybody would see the copyright statement when they navigated to the folder.

Yes and no. They'll get the list of files in the folder and then have to read the raw html unless they know how to get a web-rendered version of the page.

With a readme file, the text of the document gets displayed after the file listing -- like on the main page, where we explain the purpose of the repository.

Another option would be to add a readme file with a link to the copyright page. If you use the github pages address, you can read the document as a web page, but people would still need a link to get that view. Here's the address now that the pull request is merged: https://daisy.github.io/ebraille/experiments/copyright-statement.html

PR #152, opened 2024-02-27
Add document status to header

mattgarrish:
Because we're using the respec base profile to add DAISY's branding, we don't have access to use the built-in specification status identifiers (plus these are W3C-specific).

This pull request adds a new property called "daisyStatus" to the respec metadata to identify the status. For now, I've matched the respec pattern of it having the values "ED" for editor's draft and "WD" for working draft. (There are also "CR", "PR", and "REC" placeholder codes, but we can figure out if we need some or all of these for DAISY process later.)

When you add these codes, there's a post-processing script that will add the corresponding status designator to the header before the publication date.

The preview for the specification should provide an example of the functionality (assuming scripting works with it), so I won't add an additional preview for the tagging document.

***
<a href="https://pr-preview.s3.amazonaws.com/daisy/ebraille/pull/152.html" title="Last updated on Feb 27, 2024, 1:28 PM UTC (54904bd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/daisy/ebraille/152/056251f...54904bd.html" title="Last updated on Feb 27, 2024, 1:28 PM UTC (54904bd)">Diff</a>

### DIFF
--- a/best-practices/tagging/index.html
+++ b/best-practices/tagging/index.html
@@ -4,11 +4,13 @@
 		<meta charset="utf-8" />
 		<title>eBraille Tagging Best Practices</title>
 		<script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove"></script>
+		<script src="../../common/js/status.js" class="remove"></script>
 		<script class="remove">
 			//<![CDATA[
 			var respecConfig = {
 				shortName: 'ebraille-tag',
 				specStatus: 'base',
+				daisyStatus: 'ED',
 				latestVersion: 'https://daisy.github.io/ebraille/best-practices/tagging',
 				edDraftURI: null,
 				editors: [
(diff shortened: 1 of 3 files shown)


clapierre (approved):
Looks great Matt. 
BTW unrelated, but in the preview the Daisy Logo is a broken image. We only see the alt text, glad to see that was there ;)

mattgarrish:
> but in the preview the Daisy Logo is a broken image

Right, that's an old issue with AWS and using relative paths. The browser expects the path to resolve on the AWS server but pr-preview won't grab the common files. I guess the preview builds the page and runs the scripts before uploading it to AWS which is why the scripting worked.

@avneeshsingh should we look into getting a daisy url created that redirects to the latest logo rather than store a copy? (I got the current image at: https://daisy.org/wp-content/uploads/2019/04/daisy_high.jpg)

The benefit would be that if the logo changes in the future all the documents would update as soon as we get that redirect updated.

Issue #143, opened 2024-01-04
Braille grade indicator on language attribute


GeorgeKerscher:
## Consider using lang attribute to indicate the braille grade.

The lang attribute  should be  included in all XHTML and HTML documents. It can also be placed on other elements to indicate a change in the language. There are also provisions  for  using subtags and mechanism for extensions.

We should consider this approach as an alternative to using class attributes on span and div as described in issue number: 142

### Language tag syntax
[Language tag syntax](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang#content)

Script subtag
Optional. This subtag defines the writing system used for the language, and is always 4 characters long, with the first letter capitalized. For example, French-in-Braille is fr-

### Specification for language tags 

[                     Tags for Identifying Languages](https://datatracker.ietf.org/doc/html/rfc5646))
   This document describes the structure, content, construction, and
   semantics of language tags for use in cases where it is desirable to
   indicate the language used in an information object.  It also
   describes how to register values for use in language tags and the
   creation of user-defined extensions for private interchange.

jrbowden:
I much prefer using the lang tag to specify a braille table - no class reserved words required. The main reason we really need to note the braille table is for purposes of back translation.
I think we need to establish a scheme for identifying braille code, remembering there are cases when:
- the language is the same as the braille code, e.g. UEB contracted (English text)
- the language is different from the braille code, e.g. French in an English educational context (doesn't use full French braille code)
- there is no meaningful "language" with the braille code, e.g. music, Nemeth etc.

May I propose this ticket supersedes the examples in the Best Practice document and ticket #137 and #142 as the approach is preferable?

wfreeman-aph:
I'm scared of using lang to represent either braille codes or grades. We'd need to expand how this tag is traditionally used in order to include the various codes and it could get messy. I think we are better off using spans and divs with a class that denotes the grade and code where applicable. I get how that solution is more generalized, since we're using such generic tags but, in a way, it's also more specific since the class is going to denote exactly what it needs to for reading software to comprehend what has changed. The key, I think, is ensuring that creation and reading software use the correct terminology.

PR #160, opened 2024-04-05
Updated Tagging Best Practices

### LINKED ISSUE #3: Graphics in braille files
As a braille producer or braille reader, I want to benefit from inclusion of graphics as well as text in braille documents.

*Detail:*
Assuming I have a device capable of showing tactile graphics as well as braille, I want to make use of graphics as well as just the text in a document.
Currently, a braille document can only store the text component of the document; graphics have to be stored separately.

*Proposal:*
Tactile graphics can be stored alongside text in a braille file, so that those braille displays or embossers capable of displaying tactile graphics can incorporate them into the document.
The format for any such tactile graphics is to be discussed.

### LINKED ISSUE #4: Reflowing of text
As a braille display user or braille organisation, I want to optimally use the available braille line length when reading or embossing a braille document, which may have been produced by an organisation with a different default braille line length to my own.

*Detail:*
If my braille display or embosser has a different braille line length to the organisation who produced a document, ideally the braille in ordinary document elements should be "reflowed" to make optimal use of the available line length I have - taking into account any special formatting requirements.

What should not happen is: 
- only a portion of the line length is used (if the original document had a shorter line length to my own), or
- the document renders as full line, part line, (if the original line length is longer than my own).

Currently, braille files contain hard line breaks according to the original line length, set by the originating organisation, and these cannot easily be modified for a different target line length.

*Proposal:*
The braille file should be organised according to document structural elements, rather than by hard lines and pages fixed to a particular page size.
This would enable ordinary text elements to be flowed onto whatever line length is available.

### LINKED ISSUE #9: Internal navigation via hyperlink
As a dynamic braille display user, I want to activate links to navigate to another location within the same document.

*Detail:*
Links are a common way of traversing the web and print documents like EPUB. A common use case for links in a print document is for footnotes and endnotes. The user can activate the link and then be transferred to the location within the document that contains the text of the footnote and then activate a second link to then be transferred back to where they were and continue reading.

Another aspect of links is that they can be embedded in the text that relates to them. An example of this would be when link text is used instead of a URL on the web or when the number of a footnote represents the link to the text of the footnote.

Currently there is no means to have any kind of link within a braille document.

*Proposal:*
Links should be able to be added to the electronic braille file format. They could be modeled after links that exist on the web and in print files like EPUB so that they can be embedded within text, like link text that replaces URLs on the web or when the numbers of a footnote are the link to the footnote in an EPUB.

### LINKED ISSUE #28: Include semantic markup for multiple purposes (navigation, show/hide, backtranslation etc.)
As a reading system developer, the document should include semantic markup so that multiple use cases have the potential to be addressed.

*Detail:*
Braille users want many different options including enhanced navigation and the ability to show/hide different elements of the document. There are also braille producers that may potentially want the ability to work backwards from the braille document. All of these options require that the document has rich semantic markup that identifies the elements we want to show or hide or use for navigation. Additionally, the reading system could backtranslate and then retranslate elements if braille code switches are identified using markup as well. 

*Proposal:*
The document should include semantic markup that identifies **at least** the following:

- Emphasis changes
- Braille code changes
- Braille grade changes
- Print and braille page numbers
- Headings
- Table formatting

### LINKED ISSUE #30: Need markup that can account for spatial materials
As a transcriber, sometimes I must present materials in a way that requires multiple parts of the same line to begin in specific cells and for the material to also begin in those cells on subsequent lines.

*Detail:*
Spatial content is a normal part of braille transcription. An example would be vertically arraigned math problems. These kinds of braille materials would require, for example, that each line of the first problem begin in the 3rd cell, each line of the second problem to begin in the 8th cell, and so on. Special markup will be required to ensure that this content can be presented dynamically, rather than requiring a fixed layout.

*Proposal:*
In order to avoid requiring that these kinds of spatial materials have a fixed layout, we need markup that allows for the different parts of the spatial material to be arraigned dynamically. This could require using a container that denotes the beginning and end of the spatial materials, while also using markup specific to each part of the spatial material, so that it can be laid out in a way that respects its spatial nature but while also allowing it to be rearraigned to fit varying line lengths.

### LINKED ISSUE #47: External navigation via hyperlink
Related to #9 and #45, hyperlinks in documents could point to other documents or the same file type.
For example assume a collection of documents is delivered with a master index file. The master index could then contain a brief summary of each book and a link to open the book.

Thus we now have (at least) three possibilities for a hyperlink:
1. points somewhere in the same file in the same book - e.g. a footnote contained in the same chapter.
2. points somewhere in a different file in the same book - e.g. link from a contents to a chapter, or a "see also" which points to another chapter.
3. points to a different book - e.g. a master index opening a different book.

If needed this could be extended to allow:
4. points to a website or other external drive containing a book - e.g. book is online.
5. points to another file type entirely - e.g. opens a graphics file or other resource that the reading system is capable of rendering.

Proposal: a hyperlink could point to another (local) file.

### LINKED ISSUE #58: Transcriber-generated materials should be indicated using markup
I am a braille transcriber and I generate material that does not exist in the original print (non-braille) file. These materials vary throughout the world but can include the title page, a page for rarely encountered symbols (called a Special Symbols page in North America), and a transcriber's note page, which is a page for the rules applies to the volume and any volume-wide changes that the transcriber had to make. Experienced braille readers are more likely to skip these pages.

*Detail*
The requirements for transcriber-generated material likely vary throughout the world. Some braille regions may not use them at all. These materials include a title page, which includes the title of the book, author, the ISBN number, the transcriber agency, and the page number information related to the specific volume. Another aspect of this material is what is called a Special Symbols page in North America and this page includes information about rarely encountered characters, so the braille user can review any symbols they may not already be familiar with and understand how they are being used in this volume. Finally, another page that could be included is the Transcriber's Notes page, which includes statements about how the material was prepared, including the braille code and formatting rules used and any changes that were made so often that it is simpler to mention them on this page rather than each time the change occurs throughout the work.

*Proposal*
If this material was indicated by a kind of markup, it would empower the user to make decisions about how and whether it is produced. Some experienced braille users may opt not to reproduce this material if they were embossing their eBraille file or may choose to skip it entirely if using a refreshable display. By marking up the content, reading systems can add options that facilitate what users need most. The inclusion of this material should be optional so that it is possible for braille regions that do not use it to ignore this feature without unnecessary steps.

### LINKED ISSUE #154: Clarify for which element a class attribute is allowed
The [eBraille Tagging Best Practices](https://daisy.github.io/ebraille/best-practices/tagging/) document contains items that can be used for tagging. Each item contains the related HTML elements and the reserved class attributes.

In some cases an item has multiple HTML elements associated with it. For example the table section lists 6 related HTML elements (`table`, `tr`, `td`, `th`, etc). 

The reserved class values section defines reserved values for the `class` attribute. In the case of tables we have things like "stairstep".

In the case of tables (and probably others like lists as well) the class attribute only makes sense on some elements, i.e. the `table` element itself.

It is not really made clear that the class attribute only applies to some of the HTML elements listed.

So far this is only a problem in the table section, but if we add all the list subelements then it will be a concern there too.

### LINKED ISSUE #155: Clarify which elements are appropriate for reserved classes in Tagging Best Practices document
There is ambiguity in the [Tagging Best Practices document](https://github.com/daisy/ebraille/blob/main/best-practices/tagging/index.html). It is unclear in some situations what elements are appropriate for the reserved classes. An example of this issue would be tables, where it is unclear if the reserved classes could be applied to table, tr, th, or one of the other elements mentioned. A note should be added so it is clear which element the reserved classes can be applied to.

### LINKED ISSUE #159: Add additional contextual information to reserved class values in Tagging Best Practices document
Right now it is not clear how each reserved class value is intended to be used. The Tagging Best Practices document needs additional information about each reserved class value so that users will know what is expected.

wfree-aph:
The changes in this commit address the following issues with the Tagging Best Practices document:

- Added Matt Garrish as editor.
- Added James Bowden as author.
- Fixes #159
- Fixes #155
- Fixes #154
- Fixes #58
- Fixes #47
- Fixes #30
- Fixes #28
- Fixes #9
- Fixes #4
- Fixes #3

***
<a href="https://pr-preview.s3.amazonaws.com/daisy/ebraille/pull/160.html" title="Last updated on Apr 16, 2024, 6:09 PM UTC (321f493)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/daisy/ebraille/160/b2addfd...321f493.html" title="Last updated on Apr 16, 2024, 6:09 PM UTC (321f493)">Diff</a>

### DIFF
--- a/best-practices/tagging/index.html
+++ b/best-practices/tagging/index.html
@@ -415,7 +415,7 @@
 					<dd>
 						<p>[^ol^], [^ul^], [^dl^], [^li^]</p>
 
-						<p>Note that <code>dl</code> is preferred for <code>glossary</code>, <code>index</code>, and <code>toc</code>.</p>
+						<p>Note that <code>dl</code> is preferred for <code>glossary</code> and <code>index</code>.</p>
 					</dd>
 
 					<dt>Reserved <code>class</code> values:</dt>
--- a/best-practices/tagging/index.html
+++ b/best-practices/tagging/index.html
@@ -18,6 +18,18 @@
 						name: 'Willow Free',
 						company: 'American Printing House for the Blind',
 						url: 'https://www.aph.org/'
+					},
+					{
+						name: 'Matt Garrish',
+						company: 'DAISY Consortium',
+						url: 'https://daisy.org/'
+					}
+				],
+				authors: [
+					{
+						name: 'James Bowden',
+						company: 'Royal National Institute of Blind People',
+						url: 'https://www.rnib.org.uk/'
 					}
 				],
 				logos: [
@@ -77,9 +89,9 @@
 		
 			<section id="boxes">
 				<h4>Boxed text</h4>
-
+				<p>Boxes would be used any time that text needs to be enclosed in a full lines and/or columns of braille characters. The specific braille characters and where they appear will be determined by CSS. As an example, you might have text that is preceded and followed by a full line of the same braille character but you could also have variations or full columns of braille characters flanking the text on either side.</p>
 				<dl id="elemdef-box" class="elemdef">
-					<dt>Element(s):</dt>
+										<dt>Element(s):</dt>
 					<dd>
 						<p>[^div^]</p>
 					</dd>
@@ -88,7 +100,6 @@
 					<dd>
 						<ul class="nomark">
 							<li><code>box</code>, <code>nemeth</code>, various strings</li>
-<!--Need to know what classes to use for boxes with text, such as color or other indicators. Now sure how to define.-->
 						</ul>
 					</dd>
 				
@@ -128,7 +139,7 @@
 </section>
 			<section id="Braille grade">
 				<h4>Braille grade</h4>
-
+				<p>Braille grade refers to whether contractions are used in the braille transcription. In the United States, for example, grade 1 braille uses no contractions, while grade 2 uses any applicable contractions. Documents can be in one grade but with sections that appear in a different grade.</p>
 				<dl id="elemdef-p" class="elemdef">
 					<dt>Element(s):</dt>
 					<dd>
@@ -150,7 +161,7 @@
 			</section>
 			<section id="caption">
 				<h4>Captions</h4>
-
+				<p>Captions refer to any text in print that is associated with an image. The text is not added by the transcriber, like a description or alt text, but instead is transcribed from the print.</p>
 				<dl id="elemdef-p" class="elemdef">
 					<dt>Element(s):</dt>
 					<dd>
@@ -176,7 +187,7 @@
 			</section>
 			<section id="em">
 				<h4>Emphasis</h4>
-
+				<p>Emphasis works differently in braille than it does in print in that the emphasized characters do not themselves change but instead additional characters are added that precede, and sometimes follow, the emphasized text. The following elements will not add the braille characters themselves but will instead just track where emphasis is placed so that reading systems can potentially offer features that utilize this information.</p>
 				<dl id="elemdef-em-strong" class="elemdef">
 					<dt>Element(s):</dt>
 					<dd>
@@ -188,6 +199,11 @@
 					<dd>
 						<ul class="nomark">
 							<li><code>script</code>, <code>underline</code>, <code>custom-1</code>, <code>custom-2</code>, <code>custom-3</code>, <code>custom-4</code>, <code>custom-5</code></li>
+							<ul class="nomark">
+								<li><code>script</code> is used to denote text that is meant to mimic handwriting.</li>
+								<li><code>underline</code> is used to denote text that is underlined.</li>
+								<li><code>custom-#</code> is used in UEB to denote text that is emphasized in a way not covered by a planned type of emphasis. For example, you could use it for green highlighting since UEB has no means to denote that kind of emphasis. It could similarly be applied to other braille codes that have emphasis types not covered by the types available.</li>
+							</ul>
 						</ul>
 					</dd>
 				</dl>
@@ -200,15 +216,11 @@
 					<pre><code class="html">&lt;p>⠠⠏⠗⠑⠎⠎ ⠞⠓⠑ &lt;em class="custom-1">⠈⠼⠂⠎⠟⠥⠁⠗⠑&lt;/em> ⠃⠥⠞⠞⠕⠝⠲&lt;/p></code></pre>
 					
 				</aside>
-<p class="ednote" title="Use markup for braille-only indicators">
-  An earlier draft identified the braille-only indicators that are used for things like emphasis. This idea was not discussed with the group and has been left out. If we decide to include it, we could use spans.
-  Questions include, should we identify each kind of indicator in a unique way (bold character, bold passage, but also underline character, underline passage, etc.)-WF
-</p>
 			</section>
 
 			<section id="figures">
 				<h4>Figures</h4>
-
+				<p>[^figure^] is used to denote images. It must also include [^figcaption^] if that image is accompanied by a caption.</p>
 				<dl id="elemdef-figure" class="elemdef">
 					<dt>Element(s):</dt>
 					<dd>
@@ -245,9 +257,9 @@
 					<dt>Supported <code>role</code> values:</dt>
 					<dd>
 						<ul class="nomark">
-							<li><code>doc-noteref</code></li>
-							<li><code>doc-footnote</code></li>
-							<li><code>doc-endnotes</code></li>
+							<li><code>doc-noteref</code>- used to identify the actual symbol of the footnote or endnote. This symbol will typically appear as a superscripted number or symbol in the main body of text.</li>
+							<li><code>doc-footnote</code>- used to identify the entirity of a single footnote, so symbol and text.</li>
+							<li><code>doc-endnotes</code>- used to identify the entirity of a single endnote, so symbol and text.</li>
 						</ul>
 					</dd>
 				</dl>
@@ -272,7 +284,7 @@
 					<dt>Reserved <code>class</code> values:</dt>
 					<dd>
 						<ul class="nomark">
-							<li>Additional general formatting instructions</li>
+							<li>Additional general formatting instructions- these would be used when the transcriber has requirements for the heading that cannot otherwise be accounted for by the CSS.</li>
 						</ul>
 					</dd>
 				</dl>
@@ -342,6 +354,12 @@
 					<dd>
 						<ul class="nomark">
 							<li><code>id</code>, <code>linenum</code>, <code>prose</code>, <code>poetry</code></li>
+								<ul class = "nomark">
+									<li><code>id</code>- used to denote the actual line number in print to assist software with utilizing the information without the need for backtranslation.</li>
+									<li><code>linenum</code>- applied to the div and span to identify each as involving line numbers.</li>
+									<li><code>prose</code>- applied to the div to identify the section as involving prose text with line numbers (some braille regions have different rules for prose line numbers vs poetry line numbers).</li>
+									<li><code>poetry</code>- applied to the div to identify the section as involving poetic text with line numbers.</li>
+								</ul>
 						</ul>
 					</dd>
 				</dl>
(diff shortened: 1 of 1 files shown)


best-practices/tagging/index.html
```diff
@@ -77,9 +89,9 @@ <h3>eBraille tagging</h3>
 		
 			<section id="boxes">
 				<h4>Boxed text</h4>
-
+				<p>Boxes would be used any time that text needs to be enclosed in a full lines and/or columns of braille characters. The specific braille characters and where they appear will be determined by CSS. As an example, you might have text that is preceded and followed by a full line of the same braille character but you could also have variations or full columns of braille characters flanking the text on either side.</p>
```
mattgarrish, line 92:
```suggestion
				<p>Boxes are used any time that text needs to be enclosed in full lines and/or columns of braille characters. The specific braille characters and where they appear will be determined by CSS. As an example, you might have text that is preceded and followed by a full line of the same braille character but you could also have variations or full columns of braille characters flanking the text on either side.</p>
```

best-practices/tagging/index.html
```diff
@@ -176,7 +187,7 @@ <h4>Captions</h4>
 			</section>
 			<section id="em">
 				<h4>Emphasis</h4>
-
+				<p>Emphasis works differently in braille than it does in print in that the emphasized characters do not themselves change but instead additional characters are added that precede, and sometimes follow, the emphasized text. The following elements will not add the braille characters themselves but will instead just track where emphasis is placed so that reading systems can potentially offer features that utilize this information.</p>
```
mattgarrish, line 190:
```suggestion
				<p>Emphasis works differently in braille than it does in print. Instead of changing the appearance of emphasized characters, additional characters are added that precede, and sometimes follow, the emphasized text. The following elements will not add the braille characters themselves but will instead just track where emphasis is placed so that reading systems can potentially offer features that utilize this information.</p>
```

best-practices/tagging/index.html
```diff
@@ -200,15 +216,11 @@ <h4>Emphasis</h4>
...
-<p class="ednote" title="Use markup for braille-only indicators">
-  An earlier draft identified the braille-only indicators that are used for things like emphasis. This idea was not discussed with the group and has been left out. If we decide to include it, we could use spans.
-  Questions include, should we identify each kind of indicator in a unique way (bold character, bold passage, but also underline character, underline passage, etc.)-WF
-</p>
 			</section>
 
 			<section id="figures">
 				<h4>Figures</h4>
-
+				<p>[^figure^] is used to denote images. It must also include [^figcaption^] if that image is accompanied by a caption.</p>
```
mattgarrish, line 223: This might be too limiting a description. A figure tag could also encapsulate a table, code example, etc. Any figure with a caption, essentially.

best-practices/tagging/index.html
```diff
@@ -342,6 +354,12 @@ <h4>Line numbers</h4>
 					<dd>
 						<ul class="nomark">
 							<li><code>id</code>, <code>linenum</code>, <code>prose</code>, <code>poetry</code></li>
+								<ul class = "nomark">
+									<li><code>id</code>- used to denote the actual line number in print to assist software with utilizing the information without the need for backtranslation.</li>
```
mattgarrish, line 358: Just a general note that em dashes are usually best to offset descriptions. You can add these quickly with the entity &amp;#8212;

best-practices/tagging/index.html
```diff
@@ -401,14 +420,26 @@ <h4>Lists</h4>
 					<dt>Element(s):</dt>
 					<dd>
 						<p>[^ol^], [^ul^], [^dl^], [^li^]</p>
+
+						<p>Note that <code>ol</code> is preferred for <code>exercise</code>, <code>poetry</code>, and <code>generic</code>, while <code>dl</code> is preferred for <code>glossary</code>, <code>index</code>, and <code>toc</code>. <code>ul</code> can be used for basic lists.</p>
```
mattgarrish, line 424: Is this correct that dl is preferred for table of contents and indexes? Example 24 shows toc on an ol, for example, which is how epub also requires the table of contents be done in the navigation document. I'm not sure I can visualize how to use dd/dt pairs for these.

wfree-aph:
Thanks for this thorough review James! I've made all the changes you suggested except for the ones I've left below. Those changes are either currently reflected in issues or should have issues opened about them. They are all just "too big" for pull request feedback. The pull request can be approved even with these issues and then we can address the issues later.

--
Section 2.2:

Note: I am still uncomfortable giving "grade 1" a class of "grade-0". I still think it will inevitably lead to problems - especially as there really is a "grade 0" to some.

I think this one isn't easily resolvable immediately, and suggest we need to establish a naming convention to encompass all braille codes, see #143<https://github.com/daisy/ebraille/issues/143>.

--

Section 2.4: Emphasis.


Note: do we need to note that em is not used for code listings, sub- and superscript and transcriber notes (even though they may have special braille indicators around them)?

--

Section 2.7: headings
Note: I still think we should separate heading from header. They do very different things. So suggest insert a new section for title and header/footer. Another consideration for header/footer is they should be shortened if there is insufficient space on the (single) line - not the case for headings.

--
Section 2.8: Images (Graphics)
Note: do we need to say that alt text is mandatory?

--

Sectin 2.10: Links
"internal or external". This needs some elaboration. I can think of three kinds:

  1.  A same-page link - a link to somewhere else in the same actual HTML file.
  2.
     *   same-package link - a link to somewhere in a different file, but still part of the package as a whole.
  3.

     *   Web link: a link to a resource not on this device at all, one might say internet link.

There's a fourth type which would be to somewhere on another package still on the local device, but I think that was excluded somewhere along the line. Reckon it might be useful for indexes of content.

These three are shown in the examples.

--
Section 2.11: Lists
I'm not sure I agree that poetry is a list. I would encode it as paragraphs:

It's fleece was white as snow<br/>

...

</p>

<p class="poetr">Mary's lamb could not but grow<br/>

...

</p>



We should probably have a stanza or two somewhere in the documement.



Granted some hymns may have verse numbers, but these are as likely as not optional and may have other symbols as well as just the number, e.g. verse 1, 2, 3, *4 and 5.



There's still the question of what puts in the bullet point, list item number etc. Suggest this needs specifying, e.g.

By default, the reading system does not supply any list item symbols, akin to the CSS list-style-type: none.



I like the examples of exercise, etc, but is this a bit BANA-centric?



Regarding reserved classes, can we use the usual names of things such as none, bulleted, numbered etc?



I think a lot more work is needed here. Particularly table of contents, and we may need to split this section.



--



Section 2.12: Page numbers

Again, I think this section needs splitting: braille page numbers, print page numbers and pprint page change indiators. Actually we only need tag the print page change indicators, as is current, but probably more info would help.



"doc-pagebreak": I'm not convinced about "two contiguous". That is the normal case, yes, but sometimes there is a break in the numbering, or the numbeing resets... we still need doc-pagebreak to denote where the print page changes. e.g. the book goes from Roman to Arabic numbering, or page 66 is left intentionally blank and has no number on it...





--

Section 2.14: Preformatted



Can I suggest a couple of additions here:



One additional effect of pre is that multiple spaces are preserved, in contrast to normal behaviour of compressed.



For the example, can we add it is a couple of bars of music set out in the bar-over-bar layout? (it might help people understand why they can't read it if te don't know the musc code.)



Technicaly, we ought to add a tag to indicate the braille code switch - see braille grade above and #143.



--



Section 2.15: Tables



What abot the tabula kind of table? Sholdn't tis also be indicated as a recommendation for the reading system?



Do we need classes for things like transposed columns and rows, or a two-page spread?



Should example 31 and 32 be switched over: 31 has th tag and a class, whereas 32 only has th.



--



Note: after the end of section 2, I get an error message that

/uploads/WwJfK9/common/acknowledgements.html

cannot be found.



--------

best-practices/tagging/index.html
wfree-aph: Fixed through responding to James' notes.

best-practices/tagging/index.html
wfree-aph: Fixed through responding to James' comments.

best-practices/tagging/index.html
wfree-aph: Fixed through responding to James' comments.

best-practices/tagging/index.html
wfree-aph: Changed all hyphens being used to offset descriptions to &#8212;.

best-practices/tagging/index.html
wfree-aph: Fixed. There is no reason to use dl for toc. I don't know what I was thinking when I put it there. I was probably just confused.

wfree-aph:
I've made most of the changes that were asked and for the changes that I did not make, please make an issue. I'll merge this tomorrow afternoon, so about 18 UTC, if there is no further comment on it.

PR #161, opened 2024-04-08
Updated spec requirements

mattgarrish:
This pull request makes the following changes:

- it adds a definition of ebraille file set based on the ocf abstract container
- it adds a publication resources section;
- it adds a "status of this document" section - a sort of slimmed down version of what respec generates for w3c;
- it starts to fill in some of the basic requirements for the package document and adds some bare bones examples;
- it adds introductions to all the sections to start explaining the purpose of each

There's still a lot more to do, of course, but I don't want to grow this pull request too big. It should allow us to tackle the individual sections in more detail in later pull requests.

### DIFF
--- a/index.html
+++ b/index.html
@@ -9,6 +9,7 @@
 			//<![CDATA[
 			var respecConfig = {
 				shortName: 'ebraille',
+				daisyWG: 'eBraille Working Group',
 				specStatus: 'base',
 				daisyStatus: 'ED',
 				latestVersion: 'https://daisy.github.io/ebraille/',
@@ -55,7 +56,15 @@
 					}
 				],
 				github: 'daisy/ebraille',
-				postProcess: [addDAISYStatus]
+				postProcess: [addDAISYStatus],
+				xref: ["epub-33"],
+				localBiblio: {
+					"dpub-aria" : {
+						"title": "Digital Publishing WAI-ARIA Module",
+						"href": "https://www.w3.org/TR/dpub-aria/",
+						"publisher": "W3C"
+					}
+				}
 			};
 			// ]]>
 		</script>
@@ -67,8 +76,15 @@
 			<p>The specification defines eBraille, a digital reading format for braille publications. Unlike braille
 				formats that focus on interchanging embosser-ready braille, eBraille focuses on adapting braille for
 				reading in refreshable braille displays with different line lengths.</p>
-			<p>The eBraille format uses HTML for structuring braille content, adds requirements for including metadata
-				about a publication, and defines how to package the resulting file sets for distribution.</p>
+			<p>The eBraille format is built on an EPUB file set. It uses the EPUB package document to define the
+				metadata, resources, and reading order of a publication, and XHTML content documents to structure
+				braille content.</p>
+		</section>
+		<section id="sotd">
+			<div data-include="common/status.html" data-include-replace="true"></div>
+			<p>The eBraille Working Group is seeking input on all aspects of this document. It is particularly
+				interested in implementation experience both creating eBraille publication and creating reading systems
+				in which to read them.</p>
 		</section>
 		<section id="intro">
 			<h2>Introduction</h2>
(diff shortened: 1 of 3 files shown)


mattgarrish:
Sorry, I forgot we have pr-preview running on the tagging document. Here are the preview and diff links:

* [Preview](https://raw.githack.com/daisy/ebraille/spec/epub-file-set/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/epub-file-set/index.html)

index.html
```diff
@@ -106,21 +129,189 @@ <h3>Terminology</h3>
...
+					<li>The <a data-cite="epub-33#sec-container-metainf-container.xml"><code>container.xml</code>
+							file</a> under the <a data-cite="epub-33#sec-container-metainf"><code>META-INF</code>
+							directory</a> [[epub-33]].</li>
+				</ul>
+
+				<p>In addition, it MUST contain the following files in the publication root:</p>
+
+				<ul>
+					<li>The <a data-cite="epub-33#sec-nav">EPUB navigation document</a> [[epub-33]]. This file MUST be
+						named <code>index.html</code></li>
```
clapierre, line 241: shouldn't this be index.xhtml? (if so there are a few mentions to index.html.)

index.html
```diff
@@ -106,21 +129,189 @@ <h3>Terminology</h3>
...
 
 			<section id="metadata">
 				<h3>Metadata</h3>
 
-				<div class="ednote">
-					<p>The exact method for expressing metadata in ebraille publications will be added in a future
-						update.</p>
-				</div>
+				<p>The eBraille package document metadata MUST meet all the requirements for [[epub-33]] <a
+						data-cite="epub-33#sec-pkg-metadata">metadata</a>.</p>
```
clapierre, line 309: We should also require the metadata from the EPUB Accessibility 1.1 specification.

index.html
```diff
@@ -230,11 +464,74 @@ <h3>CSS support</h3>
...
+
+				<section id="nav-toc">
+					<h4>Table of contents</h4>
+
+					<p>For compatibility with web-based rendering, the table of contents MUST also be identified by the
+						[^/role^] attribute [[html]] value <a data-cite="dpub-aria#doc-toc"><code>doc-toc</code></a>
+						[[dpub-aria]]</p>
+
+					<aside class="example" title="Table of contents identified by the epub:type and role attributes">
+						<pre><code>&lt;nav epub:type="toc" role="doc-toc">
```
clapierre, line 509: Do we really need epub:type="toc" since it doesn't really do anything I would prefer to leave it out unless it is required by legacy reading systems but even then...

index.html
```diff
@@ -230,11 +464,74 @@ <h3>CSS support</h3>
...
+
+				<section id="page-list">
+					<h4>Page list</h4>
+
+					<p>For compatibility with web-based rendering, a page list MUST also be identified by the [^/role^]
+						attribute [[html]] value <a data-cite="dpub-aria#doc-pagelist"><code>doc-pagelist</code></a>
+						[[dpub-aria]]</p>
+
+					<aside class="example" title="Page list identified by the epub:type and role attributes">
+						<pre><code>&lt;nav epub:type="page-list" role="doc-pagelist">
```
clapierre, line 523: Same here for epub:type="page-list"

index.html
mattgarrish: index.xhtml likely won't work everywhere as a default index page on the web. You'd then have to configure the directory to accept it.

EPUB doesn't recommend extensions anymore, so using .html is fine. So far as I remember, the only thing using .xhtml does is force xhtml rendering in some browsers when opened from your file system.

index.html
mattgarrish: We haven't discussed that yet. I'm leaving that section until we get more info from the metadata folks.

index.html
mattgarrish: Yes, if the idea is to have a valid epub file set. If we drop it, then the navigation document doesn't conform.

clapierre: Oh really, that's interesting.  Wow, too bad we didn't change that to use the role="doc-toc" instead.  oh well, good to know. thanks.

index.html
clapierre: Is this also a EPUB spec requirement Matt?

mattgarrish: Same problem as the last. EPUB doesn't require the page list be identified, but it requires all nav elements in the navigation document have an epub:type attribute if they're for reading system processing.

It also ties into the question of whether we want ebraille files to work as optimally as possible in mainstream reading systems. If we move away from the epub conventions, we lose access to these features.

index.html
mattgarrish: This has been one of the roadblocks to adopting html in epub. We can't just move to aria because then all reading systems that aren't up to date would fail to load the table of contents. Same applies to moving off epub:type for the navigation document.

mattgarrish:
Are you okay with merging this @clapierre? I can't fix the epub requirement issues and the others we'll get to later.

PR #163, opened 2024-04-16
Adjust relationship to epub 3

mattgarrish:
After merging the last pull request, I noticed that I made some unintentionally strong connections to aspects of epub that we haven't settled on yet. In particular, this pull request:

- removes wording about an ebraille publication being a conforming epub publication, as that would mean an ebraille publication would have to be packaged (I thought we defined that requirement elsewhere from a publication in epub, but it's part of its definition)
- removes the requirements for the mimetype and container.xml files for now. We've only agreed on the file set, and these two files are technically a requirement of the packaging. We can add them later when we figure out if they always have to be present or if we can dynamically generate them as part of the packaging.
- makes some additional minor wording changes to be clear that the file set is compatible with epub but not exclusively epub
- clarifies that the primary entry page does not have to be in the spine

* [Preview](https://raw.githack.com/daisy/ebraille/spec/file-set-mods/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/file-set-mods/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -76,9 +76,9 @@
 			<p>The specification defines eBraille, a digital reading format for braille publications. Unlike braille
 				formats that focus on interchanging embosser-ready braille, eBraille focuses on adapting braille for
 				reading in refreshable braille displays with different line lengths.</p>
-			<p>The eBraille format is built on an EPUB file set. It uses the EPUB package document to define the
-				metadata, resources, and reading order of a publication, and XHTML content documents to structure
-				braille content.</p>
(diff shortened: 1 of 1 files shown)


wfree-aph (approved):
Good catch!

PR #165, opened 2024-04-22
Add placeholder vocabulary

mattgarrish:
As the title suggests, this is nothing more than a skeleton for formally defining the braille-specific properties. I borrowed heavily from the [epub accessibility vocabulary's](https://www.w3.org/TR/epub-a11y-11/#app-a11y-vocab) layout, but all of this can be tweaked as we go. I expect the categories also will change (or be removed) as the vocabulary is fleshed out.

[Preview](https://raw.githack.com/daisy/ebraille/vocab/placeholder/index.html)

### DIFF
--- a/vocab/index.html
+++ b/vocab/index.html
@@ -1,0 +1,218 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" lang="en-US" xml:lang="en-US">
+	<head>
+		<meta charset="utf-8" />
+		<title>eBraille Metadata Vocabulary</title>
+		<script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove"></script>
+		<script src="../common/js/status.js" class="remove"></script>
+		<script class="remove">
+			//<![CDATA[
+			var respecConfig = {
+				shortName: 'ebraille',
+				daisyWG: 'eBraille Working Group',
+				specStatus: 'base',
+				daisyStatus: 'ED',
+				latestVersion: 'https://daisy.github.io/ebraille/',
+				edDraftURI: null,
+				editors: [
+					{
+						name: 'Willow Free',
+						company: 'American Printing House for the Blind',
+						url: 'https://www.aph.org/'
+					},
+					{
+						name: 'Matt Garrish',
+						company: 'DAISY Consortium',
+						url: 'https://daisy.org/'
+					}
+				],
+				logos: [
+					{
+						src: 'common/images/daisy_high.jpg',
+						width: 112,
+						height: 100,
+						alt: 'DAISY Consortium logo',
+						url: 'https://daisy.org'
+					}
+				],
+				github: 'daisy/ebraille',
+				postProcess: [addDAISYStatus],
+				xref: ["epub-33"],
+				localBiblio: {
+					"ebraille": {
+						"title": "eBraille",
+						"href": "https://daisy.github.io/ebraille",
+						"editors": [
+							"Willow Free"
+						]
+					},
+				}
+			};
+			// ]]>
+		</script>
+		<link rel="stylesheet" href="../common/css/common.css" />
+		<style>
+			.conf-pattern {
+				margin-left: 3rem;
+				font-style: italic;
+			}
+			.varlist {
+				margin-left: 3rem;
+			}
+			pre {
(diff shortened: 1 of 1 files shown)


wfree-aph (approved):
Thank you for preparing this framework/skeleton!

clapierre (approved):
Need to put some meat on that skeleton :)

Looks good.

mattgarrish:
> Need to put some meat on that skeleton

Since you asked... I've added the properties that were in the email for the last meeting. I still don't know if the categorization of properties is helpful, or if I have them right in this first pass, but see if the new preview is better. I also added a section to describe the table fields.

vocab/index.html
```diff
@@ -160,42 +230,111 @@ <h3>abc</h3>
...
+						<td>
+							<p>Identifies the format(s) of any tactile graphics included in the work.</p>
+							<p>Repeat the property for each different format.</p>
+						</td>
+					</tr>
+					<tr>
+						<th>Allowed value(s):</th>
+						<td>
+							<code>JPG</code> | <code>PDF</code> | <code>PNG</code> | <code>SVG</code>
 						</td>
```
clapierre, line 338: Wouldn't we will want Cardinality here of zero or one?

clapierre (approved):
Looks good Matt

vocab/index.html
mattgarrish: Since you can have tactical graphics in multiple formats, you need to allow multiple instances of the property.

EPUB doesn't define a way of parsing metadata properties, so if you try to overload all the values in one string, then you open the can of worms that is parsing that back into tokens. The typical approach is to repeat a property once for each unique value (like identifiers, language, creators, access modes, etc.).

clapierre: Right, but don't we want to say that you can have 0 or more of these separate metadata properties, each of which would have one value for the type of tactile graphic?

mattgarrish: Sorry, I read your first comment as asking for zero or one.

I was going by the word doc @wfree-aph sent around which has this as a required field. That means there must be at least one. Did that change in the discussions yesterday?

clapierre: I would think that there may not be any because there may not be a tactile image of any kind in the eBRF file. @wfree-aph don't you agree?  Is the original document Matt is referencing in error?

mattgarrish: No, this is my error. It's in the required section of the document but is only conditionally required when tactileGraphic is set to true. My reading skills are a bit diminished this week...☹️

I've fixed the definition.

clapierre (approved):
Much better!

egli:
There was some discussion at the meeting on Tuesday where we re-evaluated some of the metadata. I don't see the notes in the wiki, but from my recollection I think that volumes only make sense in a paginated environment, so they might get dropped

mattgarrish:
> I think that volumes only make sense in a paginated environment, so they might get dropped

Ya, that was my concern in adding properties too soon, but at the same time this is just a draft and I'm sure it will change many times before we're done. I can comment that one out for now if it makes more sense.

The document also only covers the properties that are required and recommended right now, but I expect there will be more optional braille-specific properties that will still need to be formally defined in the vocabulary.

I'm also not sure if it makes sense to keep it separate from the main specification. We went the other way in the last epub revision and merged all the package document vocabularies back into main specification. It's awkward to have normative requirements outside of the main specification, and it makes finding information more complicated when you have to go to multiple sources.

But again, this is why I only consider this an early draft right now.

egli:
> > I think that volumes only make sense in a paginated environment, so they might get dropped
> 
> I can comment that one out for now if it makes more sense.

Thanks

> It's awkward to have normative requirements outside of the main specification, and it makes finding information more complicated when you have to go to multiple sources.

Ah, yes, good point, haven't thought of that. Might make sense to merge the two.

mattgarrish:
> Might make sense to merge the two.

For early development, it's probably easiest to have it as a separate document so we can work on it in isolation, but once it gets more mature we'll definitely need to look at whether having it separate makes sense.

I'm going to merge it now and then we can start opening pull requests to further flesh it out.

vocab/index.html
wfree-aph: Sorry for my lateness on this discussion but glad it was worked out.

PR #174, opened 2024-05-17
Add restriction on declaring fixed layout properties

mattgarrish:
This pull request explicitly bans the fixed layout properties from being specified. I was going to blanket ban all of them, since the default layout is always reflowable, but there may be some good reason why someone wants to explicitly declare their publication is reflowable.

* [Preview](https://raw.githack.com/daisy/ebraille/spec/no-fxl/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/no-fxl/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -479,6 +479,21 @@
 				</div>
 			</section>
 		</section>
+		<section id="ebrl-rendering-control">
+			<h2>Layout rendering control</h2>
+
+			<p>eBraille does not support <a data-cite="epub-33#sec-fixed-layouts">fixed layouts</a> as defined in
+				[[epub-33]]. Consequently, [=eBraille publications=]:</p>
+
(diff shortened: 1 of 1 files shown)


wfree-aph (approved):
Looks great! Your ability to write using the voice required for these kinds of documents is really impressive to me.

mattgarrish:
> Your ability to write using the voice required for these kinds of documents is really impressive to me.

Thanks, but I worry about the day when I'll lose my ability to think or speak in anything but spec-ese! 😄

PR #175, opened 2024-05-17
Add best practices and accessibility sections

mattgarrish:
I've moved the mention of the tagging best practices from a note in the introduction to content documents into its own section. I've also added a link to the future styling guide although I realize it won't actually work.

I also made a start at an accessibility section to push people to follow the epub accessibility specification, as much as it can apply to a braille document. (Making the metadata mandatory could be a later change.)

* [Preview](https://raw.githack.com/daisy/ebraille/spec/authoring/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/authoring/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -449,18 +449,16 @@
 			<section id="ecd-intro" class="informative">
 				<h3>Introduction</h3>
 
-				<p>eBraille only supports XHTML for authoring the content of an <a>eBraille publication</a>. Unlike EPUB
-					3, SVG is not supported in the [=spine=] but can be embedded in XHTML documents.</p>
+				<p>[=eBraille content documents=] define the markup and presentation of the content of an [=eBraille
+					publication=]. An eBraille content document is encoded using <a data-cite="html#">XML syntax</a> of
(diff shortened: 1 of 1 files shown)


clapierre (approved):
Sounds familiar ;) (aka Fixed Layout not being able to meet all WCAG requirements.)

Nice work.

mattgarrish:
It's more that they're not applicable because they don't make a lot of sense. What does it mean to require braille characters zoom to 200% without loss of functionality, for example? Can tactile images fail image contrast requirements? Does braille text have to meet the text spacing requirements?

I don't know that we can ignore the success criteria that aren't useful and still claim conformance, which is where the whole optimized publication thing came in. We probably have to define a stripped-down standard specially for ebraille.

bertfrees (approved):

wfree-aph (approved):
Looks great! Thanks for adding these sections.

PR #176, opened 2024-05-19
Add media overlays section

mattgarrish:
There's not a lot to say about media overlays beyond what epub already defines so I decided to give a brief recap of them as part of mentioning that they're supported. If people are okay with this approach, I might use it as a template for fleshing out some of the package document sections. Otherwise, we're kind of assuming everyone reading the specification will be fully familiar with the inner workings of epub.

* [Preview](https://raw.githack.com/daisy/ebraille/spec/mo/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/mo/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -573,9 +573,9 @@
 			<h2>Media overlays</h2>
 
 			<p>eBraille publications support <a data-cite="epub-33##sec-media-overlays">media overlays</a> as defined in
-				[[epub-33]]. Media overlays allow prerecorded audio playback to be synchronized with the content of
-				[=eBraille content documents=], allowing users to switch between reading braille and listening to
-				auditory playback.</p>
+				[[epub-33]]. Media overlays allow prerecorded audio to be synchronized with the content of [=eBraille
+				content documents=], allowing users to switch between reading braille and listening to auditory
+				playback.</p>
 
 			<p>A media overlay document MAY be associated with an [=eBraille content document=] by adding a
 					<code>media-overlays</code> attribute to its package document manifest entry.</p>
--- a/index.html
+++ b/index.html
@@ -574,8 +574,8 @@
 
 			<p>eBraille publications support <a data-cite="epub-33##sec-media-overlays">media overlays</a> as defined in
 				[[epub-33]]. Media overlays allow prerecorded audio to be synchronized with the content of [=eBraille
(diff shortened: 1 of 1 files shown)


avneeshsingh:
It is important to mention Media Overlays in Ebraille document because some DAISY members are hoping that they would be synchronizing audio with braille in future.
So, it is important to have this section as reminder or confirmation that Media Overlays is supported, without repeating whole MO spec here.

GeorgeKerscher:
Hi,

Looks good, but I have one question. Where it says:
allowing users to switch between reading braille and listening to auditory playback.

George thinks, I would envision it would enable the simultaneous reading of braille and listening to audio content.

mattgarrish:
> the simultaneous reading of braille and listening to audio content

This is what I don't fully understand about synchronizing audio with braille. How does the display work on the refreshable device? If you synchronize to the word, does the display have to render one word at a time as quickly as it's being narrated? Or does it show as much text as it can and only refreshes once you pass what has been rendered? (i.e., there's no indication what word is being read and you just have to follow along?)

Or, another scenario, if you're synchronized by paragraph, if you manually refresh the braille display to keep up as the audio passes what is displayed, would that take you out of playback mode? Or can the user do whatever they want at the display level during playback and the reading system just forces them ahead each time a new synchronization point is reached?

clapierre:
I could imagine multiple ways that a reading system could synchronize audio with Braille in the in the future if the spec allows it.
For example on displays that have multi height pins. the word being spoken could be at the highest pin level and the surrounding text at the lower level.  Or there could be some cursor feature following the words etc.

clapierre (approved):

wfree-aph (approved):
Great addition!

mattgarrish:
I've modified the sentence to account for all possibilities:

> allowing users to switch between reading braille, listening to auditory playback, or reading along with the narration.

clapierre (approved):
Yup, looks better, thanks Matt

bertfrees (approved):

PR #177, opened 2024-05-21
Add ebraille reading system section

mattgarrish:
Here's a first stab at reading system requirements. I went through the epub reading systems spec and listed out all the differences I could find in terms of support.

I'm less sure about the section on web-hosted publications, but leaving it in for now.

One question is do we want to mandate support for PDF for tactile images, should it be a recommendation, or do we need to mention it at all?

* [Preview](https://raw.githack.com/daisy/ebraille/spec/rs/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/rs/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -763,18 +763,18 @@
 						>requirements for EPUB reading systems</a> [[epub-rs-33]], with the following differences:</p>
 
 				<ul>
-					<li>If it supports web access, the reading system SHOULD allow users to download unpackaged
-						[=eBraille publications=] hosted on the web by specifying the path to the package document.</li>
-					<li>It MUST support PDF for tactile graphics.</li>
 					<li>It MUST NOT support <a data-cite="epub-rs-33#sec-epub-rs-conf-remote-res">remote resources</a>
 						[[epub-rs-33]].</li>
-					<li>As remote resources, forms, and scripting are not supported, it MUST NOT allow eBraille
-						publications to have <a data-cite="epub-rs-33#sec-epub-rs-network-access">network access</a>
+					<li>As remote resources, forms, and scripting are not supported, it MUST NOT allow [=eBraille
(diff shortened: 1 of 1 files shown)


wfree-aph (commented):
This looks good to me. As far as PDF support, one question I would ask is what about a reading system that offloads PDF support to another app? Like what if you read the braille using one app and view the PDF using another app? Would that mean the reading system was not valid? It seems to me that the ability to recognize PDFs and trigger them to open in the other app would count as "support" but please let me know if I am wrong. If that's the case, I think requiring PDF support is the correct path because of the problems that making it optional could cause.

mattgarrish:
> what about a reading system that offloads PDF support to another app?

That's not how EPUB would define support, but that's not really an issue here. We just need to define that as an acceptable behaviour, so maybe I'll move that bullet into a new subsection (it's technically an addition to EPUB's requirements whereas all the rest are removing features).

mattgarrish:
Rather than add new subsection, I just took the first two bullets and made them separate paragraphs after the list. The question I can't answer is whether there are technical issues that would prevent a pdf from opening in another app, but that's the sort of feedback we'll have to wait on implementers for.

I also added another bullet about not supporting non-xhtml documents in the spine. Not supporting manifest fallbacks doesn't answer what to do if someone were to put another format in the spine. I've basically left it to the reading system to either reject such a publication outright or to ignore any spine items that reference foreign resources.

bertfrees (approved):

mattgarrish:
I noticed I had put a couple of reading system restrictions in the authoring sections, so I moved the one about not resolving path absolute url strings to the new section and removed the redundant one about not supporting scripting and form submissions.

PR #178, opened 2024-05-23
Explain manifest and spine in more detail

mattgarrish:
Similar to the media overlays PR, I've expanded the sections on the package document manifest and spine to explain a bit more what they are for, what is required of them, and to give some examples. Additional links are provided to the fuller epub definitions.

* [Preview](https://raw.githack.com/daisy/ebraille/spec/pkg-doc/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/pkg-doc/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -437,11 +437,16 @@
 						>navigation document</a> and when SVG tactile graphics are embedded in the resource.</p>
 
 				<aside class="example" title="Manifest entry for an eBraille content document with embedded SVG">
+					<p>In this example, the <code>chapter02.html</code> file contains one or more SVG images. These
+						images could be embedded directly in the HTML document using the <code>svg</code> element or
+						they could be separate resources referenced from <code>img</code> or <code>object</code>
+						elements (in which case the images would also be listed in the manifest).</p>
 					<pre><code>&lt;package &#8230;>
    &#8230;
    &lt;manifest>
+      &#8230;
       &lt;item id="c02"
-            href="eBraille/chapter02.xhtml"
+            href="eBraille/chapter02.html"
             media-type="application/xhtml+xml"
             properties="svg"
       &#8230;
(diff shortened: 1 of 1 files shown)


clapierre (approved):
Looks good Matt.
Question to be complete with the example regarding the SVG in the package document shouldn't you also include the image itself in the images folder?
ie.
`    <item id="tactile-image1.svg" href="Images/tactile-image1.svg" media-type="image/svg+xml"/>`

mattgarrish:
> shouldn't you also include the image itself in the images folder?

I try to keep the examples as minimal as possible to avoid confusion. There wouldn't be a connection between the entries absent html markup to show the svg referenced from an img or object tag. And that'd be making a massive example just to show a single attribute value.

But maybe what I can do is add a description to the example to explain what it means that it has embedded svg.

GeorgeKerscher:
Looks good.

clapierre (commented):
why did you change from xhtml to html?  I thought we wanted these embedded files to be xhtml so that it would be in alignment with EPUBs.  Did we make a change?

mattgarrish:
> why did you change from xhtml to html?

To be consistent with using .html for the navigation document/primary entry page.

The extension doesn't define how the contents are rendered, so .html can be used for both syntaxes. Browsers use the media type the file is served as to determine which parser to use, but in the case of epub and ebraille it's irrelevant since the reading system should always be serving the files as xhtml.

Since we talk about serving ebraille file sets on the web, it may be worth noting the need for the correct media type in another section, maybe in the reading system section. Let's deal with this in another pull request, though.

mattgarrish:
For the record, epub doesn't require .xhtml extensions, either. It used to be a recommendation, but that got tossed in 3.3 along with a lot of other file extension requirements..

bertfrees (approved):

PR #181, opened 2024-05-30
Explain navigation elements in more detail

mattgarrish:
Not much to add to the title. Expands each of the nav sections to include more detail about their use.

* [Preview](https://raw.githack.com/daisy/ebraille/spec/nav/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/nav/index.html)


clapierre:
Matt, I would suggest including a `<h1>` for each of these (TOC, Page List, Landmarks, etc.) and have the `<nav>` point to it using aria-labelled by as each Nav must have a unique heading which is fine if there is only one `<nav>` in the file but when there are multiple its better to show the explicit headings for each.

```
Eg.
<nav aria-labelledby="toc" ...>
   <h1 id="toc">Table of Contents</h1>
   ...

<nav aria-labelledby="pagelist"...>
   <h1 id="pagelist>Page List</h1>
   ...

etc.

```

mattgarrish:
Ah, right, I wasn't focused on headings and they may need to be rendered for web deployment.

But headings are only required if there are multiple of the same role, not the same element, and you can't have more than one of each of these nav types, so a heading is not absolutely necessary from a machine processing point of view. If you only have these three, they'd be identified as the table of contents, page list, and probably a generic navigation region.

clapierre (approved):

mattgarrish:
Actually, this doesn't seem right. The headings should be in braille if they're for web rendering and there should probably be aria-label attributes with the plain language names for AT.

bertfrees (approved):

wfree-aph (approved):
Looks good and thanks for adding it!

mattgarrish:
I've updated the page list to require the title attribute. Do you think we need to say that the navigation document should be encoded in braille or does that stand to reason given the nature of the format? (I'm assuming we shouldn't need to say anything more given the very nature of the format.)

Also, can you double-check the examples for me, @wfree-aph? I did a character-to-character translation so may have got things wrong, or at least suboptimal.

wfree-aph (commented):
Thanks for drawing my attention to the braille. It was mostly fine, but I did spot one issue in my second (closer) review.

Line 940- remove ⠠ from before ⠉⠕⠝⠞⠑⠝⠞⠎ as that is the capital indicator and is capitalizing the 'c' in contents. The print example has a lowercase 'c', so if you make that uppercase, you can keep the ⠠

Issue #7, opened 2022-08-24
Synchronized switching between braille display and TTS audio


ManfredMuchenberger:
As a screenreader user, I want to be able to read an e-books on a braille display and after a while switch to audio and continue reading at the next paragraph of the e-book using the synthetic voice.  

_Detail:_
Currently, e-books only contain an ink print version of the text.
It is easy to listen to the e-book using the synthetic voice of the Screenreader. But when switching to reading on the braille display that is also connected to the screenreader you don’t get good quality braille.

_Proposal:_
Using a digital text file format that contains the ink print version of the text as well as a pre-transcribed braille version of the text in Unicode braille characters that is synchronized with the ink print version.

avneeshsingh:
What should be the priority: High, Medium or low?
My view is that it is not as high priority as navigation, reflowability etc., should we mark it medium or low priority?

ManfredMuchenberger:
In my opinion this is quite important for users. Otherwise they have to buy/download two differnet filesets if they want to use audio and braille display. And we would have to provide both filesets. Also then there would be no synchronization. So I would at least assign High.

wfreeman-aph:
This is a problem we need to prioritize soon. If we create a braille-first file like we said in the problem statement, then that will potentially lead to some TTS audio issues. The problem statement also includes this statement, "although we will discuss the possibility of including other information types and multimodal reading". So we'll need some consensus. Having braille and Latin script in the same file could be burdensome for braille producers and software but not having both could be burdensome for libraries and distributors. Alternatively, we could put the burden on screen readers and ask them to handle the problem of reading braille text. This is one we may want to discuss in the meeting.

mrhunsaker:
The issue I see with TTS and a braille-first file format is that back-translation of braille (I speak with experience primarily in UEB ) is imperfect, and thus the TTS may read incorrect information. I could envision a system whereby the original print text was somewhat associated with the corresponding braille so the TTS could follow along. However, this seems more of a print-first than braille-first solution.

mhorspool:
My view is that this is out of scope for this standard. Most screen readers can handle braille translation sufficiently for this not to be needed in the vast majority of cases. For those specific cases where it is needed, we should encourage existing synchronised audio/text standards (e.g. DAISY) to permit braille instead of text.

wfreeman-aph:
[...]

PR #182, opened 2024-05-31
Add ebraille publication conformance and expand resource requirements

mattgarrish:
Filling in a few more missing pieces. This give the high-level requirements for an ebraille publication. I've also filled out the publication resources section with missing detail. Even if we aren't changing a lot from epub, we shouldn't just omit mentioning the requirements exist.

The only changes this makes from EPUB is that it only allows one rendition of the content. I can't imagine we want to get into selecting from different renditions in different braille codes.

It also requires utf-8 encoding for xml. EPUB can only recommend it because we allowed utf-16 for a long time before the web started shifting away from it, but no need to allow it here.

* [Preview](https://raw.githack.com/daisy/ebraille/spec/pub-req/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/pub-req/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -177,7 +177,7 @@
 				<p>The [=eBraille file set=] is designed to be compatible with EPUB 3 [[epub-33]] and adapt to changes
 					to that standard. Distribution and consumption on mainstream [=EPUB reading systems=], however, is
 					not a primary goal of this format. This specification introduces some additional requirements and
-					features beyond those defined in EPUB 3, and it is not expect that mainstream reading systems will
+					features beyond those defined in EPUB 3, and it is not expected that mainstream reading systems will
 					adapt to these modifications. Consequently, an eBraille publication may not render exactly as
 					intended outside of eBraille reading systems.</p>
 
@@ -208,6 +208,31 @@
 				</ul>
 			</section>
 		</section>
+		<section id="ebrl-pub">
+			<h2>eBraille publication conformance</h2>
+
+			<p>A conforming [=eBraille publication=]:</p>
+
+			<ul>
+				<li>
+					<p>MUST define exactly one rendering of its content as follows:</p>
+					<ul>
+						<li>It MUST contain a package document that conforms to <a href="#ebrl-package-doc"></a> and
+							meet all publication resource requirements for the package document.</li>
+						<li>It MUST contain a primary entry page that conforms to <a href="#ebrl-nav"></a>.</li>
+					</ul>
+				</li>
+				<li>SHOULD conform to the accessibility requirements defined in <a href="#ebrl-a11y"></a>.</li>
+			</ul>
+
+			<p>In addition, all publication resources MUST adhere to the requirements in <a href="#ebrl-resources"
+				></a>.</p>
+
+			<p>If an eBraille publication is packaged for distribution, it MUST adhere to the packaging requirements in
+					<a href="#ebrl-packaging"></a>.</p>
+
+			<p>The rest of this specification covers specific conformance details.</p>
+		</section>
 		<section id="ebrl-resources">
 			<h2>Publication resources</h2>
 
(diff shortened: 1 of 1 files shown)


clapierre (approved):
nice job Matt.

ManfredMuchenberger:
Thank you Matt - it looks like you did a lot of work for this project!!!

Regarding the renditions I still think, that we will need to meet the use case where people want to switch between braille display and TTS audio in a synchronized way. At least we at SBS would prefer to be able to combine E-Braille with our current E-Books in EPUB format, so that we can distribute them as a package and the user can switch between braille display and TTS audio.
https://github.com/daisy/ebraille/issues/7

Therefore I think we should allow at least one Braille plus one ink print rendition of the text - if possible.

I know, that this use case is not in the MVP but I guess we are already adding other precautions for other use cases that are not in the MVP in a strict way.

By the way - I think I somehow last track of the EPUB standards. I got the impression, that multiple renditions have been droped with EPUB 3.2 . But now multiple rendition are allowed in EPUB 3.3 again right?

mattgarrish:
> we will need to meet the use case where people want to switch between braille display and TTS audio in a synchronized way

Hi Manfred,

That's probably going to prove to be difficult to do, but I'm fine allowing more than one rendition if there's a case for it. I've just been trying to scale back epub's unused features.

The basic attributes that allow selection between multiple renditions have no real support in epub, but that's largely because there hasn't been a use case for them. Maybe dedicated ebraille reading systems will take up the challenge here, though. What you're going to find time consuming is trying to create a mapping document using EPUB CFIs so that users can get to the same location they were at. I don't know of any tool that can create such a thing, and it's not something you can realistically do manually with any granularity.

> I got the impression, that multiple renditions have been droped with EPUB 3.2 . But now multiple rendition are allowed in EPUB 3.3 again right?

No, it wasn't dropped in 3.2 in the sense of being disallowed. When we brought epub into W3C, we stopped including normative references to all the other specifications we developed in IDPF. That didn't make them invalid to use, but it avoided us having to port all those documents into W3C at the same time as the main standard.

In 3.3, we started moving more directly away from multiple renditions in epub. It's too late in epub's life to remove the functionality, but the core spec no longer talks about an epub containing multiple renditions, even though it's allowed. We instead created a [working group note](https://www.w3.org/TR/epub-multi-rend-11/) out of the old IDPF specification and added in the parts from the epub specification so everything about multiple renditions is now covered in that one note.

Or, more simply, multiple renditions are no more or less allowed now than at any time in the past, but I expect their future in mainstream epubs is likely to only ever be theoretical, and will disappear whenever a successor to epub 3 is developed.

mattgarrish:
> but I'm fine allowing more than one rendition if there's a case for it

Actually, when I think about this some more, multiple renditions for ebraille is complicated by the required naming and location of the package document and primary entry page.

The only way it works is if subsequent renditions are exempt from these two requirements. I'll at least change the wording from "exactly one rendering" to "at least one rendering" for now to match epub, and then in a subsequent pull request we can try and figure out how the exemption will work for additional renditions.

ManfredMuchenberger:
Thank you Matt for the clarification. This is very helpfull!
I am fine with your new wording.

Then we can still try to find out a way to make this work at a later stage. I guess that a lot also depends on how granular the synchronization between renditions needs to be. 
I do have a sample of an EPUB with two renditions (one rendition with UNICODE Braille and one rendition with ink print) that Bert Frees did some years ago.
Our plan then was, to work together with Dolphin to see how the switching from one to the other rendition could be done.
There is a file included called "renditionMapping.html" but this is way above my skill level to understand how this was supposed to be working.

mattgarrish:
> There is a file included called "renditionMapping.html" but this is way above my skill level to understand how this was supposed to be working.

Yes, that would be the rendition mapping document, I expect. It uses CFIs to point to the corresponding locations in each rendition. The reading system has to determine the closest mapping point to where the user is and then can jump them to the same spot in the other rendition when they want to switch.

You can do very simplistic mapping documents, like only to chapter headings, but then the user has to find their way back to where they left off in the other document. The farther away they are from where they switched, the more annoying it becomes.

It's as you get more granular to make the mapping better that it becomes time consuming to do if you have to do it manually. If the braille and print markup structures are similar, though, it might be possible to automate the creation of the documents. But those are the challenges you'll have to figure out.

bertfrees:
> If the braille and print markup structures are similar, though, it might be possible to automate the creation of the documents.

The braille rendition and renditionMapping document are automatically generated from the default rendition. This is easy to do because the markup of the two renditions is indeed identical. I don't remember if we were ever able to test whether it actually works though.

mattgarrish:
> The braille rendition and renditionMapping document are automatically generated from the default rendition.

Okay, I was thinking this was a request to bundle the print publication with the braille rendition, in which case the mapping would probably not line up nicely, but if it's just a difference between braille characters or not then that would be much simpler to pull off.

Like I said above, though, it's largely dead tech in epub as publishers just aren't interested in bundling renditions. If they produce a different rendition, they publish it as a separate epub. Any support is likely going to have to come from ebraille reading systems.

PR #183, opened 2024-06-03
Add exemption for scripting in primary entry page

mattgarrish:
This will only allow script elements if the primary entry page is not listed in the spine, so not reachable by users in a dedicated ebraille reading system.

* [Preview](https://raw.githack.com/daisy/ebraille/spec/pep-scripting/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/pep-scripting/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -605,8 +605,8 @@
 
 					<div class="note">
 						<p>An exemption is made to the scripting requirement for the <a href="#ebrl-nav">primary entry
-								page</a> so long as it is not listed in the spine. This exemption is only to allow
-							[=eBraille creators=] to add a user interface for web deployment.</p>
+								page</a> so long as it is not listed in the spine. The exemption is to allow publishers
+							to add a user interface for web deployment.</p>
 					</div>
 				</section>
 			</section>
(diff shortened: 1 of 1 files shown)


GeorgeKerscher:
Looks fine; very simple.

clapierre (approved):

bertfrees:
What would be an example of "better enable reading of their publication in browsers"?

mattgarrish:
Creating a UI that lets you read the publication without leaving the primary entry page, while increasing the usability of the document.

Instead of the rudimentary experience of having to click through to each content document or heading from the table of contents, you could, for example, write an iframe to load each document and add previous and next buttons. You could script a means of jumping to specific print pages rather than have a huge page list rendered after the table of contents. Things like that.

I don't know if it takes off in practice, but it carves out an opening to experiment on building such an interface.

PR #185, opened 2024-06-04
Add section on multiple renditions

mattgarrish:
This new section exempts subsequent renditions from the file naming and location requirements. It also requires that the default rendition be tactile.

I've also added a note that the technology is not considered stable as, even though it's valid to use, it's not supported or been reviewed in the same way as epub 3 has.

* [Preview](https://raw.githack.com/daisy/ebraille/spec/multi-rend/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/multi-rend/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -363,6 +363,28 @@
 				<p>For this reason, the eBraille file set MUST NOT include file references that use path-absolute-URL
 					strings.</p>
 			</section>
+
+			<section id="multiple-renditions">
+				<h3>Multiple renditions</h3>
+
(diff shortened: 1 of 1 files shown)


clapierre (approved):

ManfredMuchenberger:
thank you Matt!

bertfrees (approved):

PR #187, opened 2024-06-07
Updates to intro sections

mattgarrish:
This pull request makes the following changes:

- it adds a couple of paragraphs to the end of the relationship to epub 3 section to explain how that also exposes us to the evolving standards that epub references (the html "living standard", unversioned svg, etc.).
- it adds an authoring shorthands section to explain that we sometimes use epub:type and dc: elements without showing their namespace declarations. I noticed we were guilty of this for epub:type already and will probably do the same once we flesh out the metadata

Otherwise, I've just rearranged the introduction a bit as it was getting a bit incoherent. The first three sections are now the descriptive ones - the overview, relationship to epub 3, and future directions - followed by the technical ones - terminology, conformance, and authoring shorthands.

* [Preview](https://raw.githack.com/daisy/ebraille/spec/intro-sections/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/intro-sections/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -107,18 +107,68 @@
 					packaging. And as web-compatible file sets, they are adaptable to future changes to publishing
 					formats.</p>
 			</section>
-
-			<section id="conformance"></section>
-
+			
+			<section id="relationship-epub3" class="informative">
+				<h3>Relationship to EPUB 3</h3>
+				
+				<p>The [=eBraille file set=] is designed to be compatible with EPUB 3 [[epub-33]] and adapt to changes
+					to that standard. Distribution and consumption on mainstream [=EPUB reading systems=], however, is
+					not a primary goal of this format. This specification introduces some additional requirements and
+					features beyond those defined in EPUB 3, and it is not expected that mainstream reading systems will
+					adapt to these modifications. Consequently, an eBraille publication may not render exactly as
+					intended outside of eBraille reading systems.</p>
+				
+				<p>The primary difference between the eBraille format and EPUB 3 is eBraille publications do not have to
+					be packaged and distributed in an [=EPUB container=] [[epub-33]]. An [=eBraille publication=] will
+					be easy to package in an EPUB container when needed, however.</p>
+				
+				<p>Compatibility with EPUB 3 also means that eBraille relies on the same <a
+					data-cite="epub-33#sec-intro-relations">underlying technologies</a> [[epub-33]]. Many of these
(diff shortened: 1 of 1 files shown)


clapierre (approved):

bertfrees (approved):

index.html
```diff
@@ -107,18 +107,68 @@ <h3>Overview</h3>
...
+					unsupported or are found to have serious security or privacy flaws.</p>
+			</section>
+			
+			<section id="future-dir" class="informative">
+				<h3>Future directions</h3>
+				
+				<p>The first version of the eBraille format is focused on creating a reading experience that is
+					minimally feature complete for the majority of braille publications. It is not expected that this
+					version will handle every unique formatting requirement of every publication type, but future drafts
+					and versions of this specification will focus on making the format more feature complete.</p>
```
bertfrees, line 147: Should we also say here that the focus lies primary on electronic
reading systems (more specifically refreshable braille displays), and
less on embossing tools? I noticed we don't mention embossers at all
yet. I think the use case deserves to be mentioned, but we should make
it clear that it is not the primary focus in the first version.

I think I would also like to clarify somewhere that:

1. It is not expected that authors always provide complete CSS style
   sheets. If a publication does not have any style sheets embedded,
   it does not mean it will not be formatted at all, or that it can not be
   formatted correctly. It is expected that reading systems apply some
   default styles that make sense for the current locale.

2. Reading systems may apply styling through other means than CSS.

3. We explain why we don't define any CSS extensions yet. In addition
   we could also explain why we intentionally omit certain standard
   CSS features, especially those that are not well supported in
   browsers, from our guidelines. The reason is that we want to wait
   and see which technologies are chosen by implementors and how they
   approach certain formatting requirements.

   E.g. CSS has a "Paged Media" module, but it is not well supported,
   so it does not make much sense to use this module in our best
   practices if it turns out to be very difficult to support by
   implementors.

   Another reason for keeping the guidelines minimal in the first version
   is that it remains to be seen how the majority of the devices will
   work: will they render the document in a "paged" way (breaking it up
   into discrete pages), or a "scrolling" way (like in a web browser).

PR #190, opened 2024-06-18
Move the metadata vocabulary into the specification

mattgarrish:
It's awkward to have normative requirements on the use of the properties defined in a separate document. It's also easy to forget about them when it comes time to update, and more work for authors to find the information.

This pull request moves the vocabulary back into the specification similar to how the epub specifications define vocabularies.

* [Preview](https://raw.githack.com/daisy/ebraille/spec/merge-vocab/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/merge-vocab/index.html)


clapierre (approved):
Makes sense Matt.

Issue #189, opened 2024-06-14
Add section on CSS media queries and CSS restictions


bertfrees:
@wfree-aph @avneeshsingh @mattgarrish @clapierre @GeorgeKerscher

This is not a PR yet. I first want to summarize what I'm about to write, ask whether it is in line with our previous discussion, and ask where exactly you think it should be put: in the formal specification, in a note, in the best practices?

### CSS restrictions

Rather than restrictions, these could also be CSS features that authors may use but reading systems should ignore. In any case, I think these should be mentioned somewhere:

- Authors must use the relative length units `em` and `ch` for all lengths. There is really no reason to use absolute lengths, especially if the author has no control over the font (see next). 
- Authors must not use font related properties. This is needed to be able to reliably use `em` and `ch` heights/widths in media queries (see below). There is no reason for the author to control the shape and dimensions of the braille dots anyway. That is a reading system/embosser thing.

### CSS media queries

- When the target medium of a style sheet is a refreshable braille display (the main use case), the author should not use any media features that discriminate between a braille display and a screen. This comes down to not using at least [`screen`](https://www.w3.org/TR/mediaqueries-4/#valdef-media-screen), [`braille`](https://www.w3.org/TR/mediaqueries-4/#valdef-media-braille) and [`grid`](https://www.w3.org/TR/mediaqueries-4/#grid)[^1]. eBraille reading systems of type braille display must match either `braille` or `screen`, and should ideally match both.

   The logical consequence of this is that it is not possible to discriminate between visual and tactile access modes, in other words this decision forces rendering on a screen and on a braille display that are otherwise comparable, to be the same, and the CSS to have the same restrictions.

   Note that we don't recommend using `braille`, nor `(grid)`, because there are currently no known implementations that match these, so requiring eBraille reading systems to match them could be problematic. Another reason to not recommend `braille` is because it is expected to become deprecated. 

   We don't recommend using `screen` because we want to be forward-compatible with a possible future version of the specification that supports differences in styling for screens vs. braille displays, might that become a requirement.

- When the target medium of a style sheet is an embossed document, the author should use `embossed`.

   Alternatively, because `embossed` is expected to become deprecated, we could recommend using <code>(grid) and (<a href="https://www.w3.org/TR/mediaqueries-4/#descdef-media-overflow-block">overflow-block: paged</a>) and not (<a href="https://www.w3.org/TR/mediaqueries-4/#descdef-media-update">update</a></code>) instead.

[...]

PR #192, opened 2024-06-20
Expand CSS section and add something on media queries

bertfrees:
Fixes https://github.com/daisy/ebraille/issues/189

@mattgarrish I need you to check that there's not too big of a break in style with what you have written so far, and also that I haven't written anything that a native English speaker wouldn't write.

@wfree-aph, @GeorgeKerscher, @clapierre, @avneeshsingh

* [Preview](https://raw.githack.com/daisy/ebraille/css/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/css/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -92,7 +92,7 @@
 
 				<p>In order to maintain close alignment with mainstream publishing formats, and simplify transcription
 					of these works, the eBraille format is built on an EPUB 3-compatible file set &#8212; it
-					incorporates XHTML documents, CSS, images, audio, and video. This means that an [=ebraille
+					incorporates XHTML documents, CSS, images, audio, and video. This means that an [=eBraille
 					publication=] shares its technology with the web, making it compatible for reading in standard web
 					browsers.</p>
 
@@ -157,6 +157,9 @@
 					<li>CSS extensions &#8212; the first version of eBraille uses standard CSS properties to format
 						braille content. There may be a need to extend CSS in the future to handle more complex
 						formatting cases, which will also require defining reading system implementations.</li>
+					<li>embossing &#8212; this version of eBraille is focused on consuming [=eBraille publications=] on
+						electronic reading systems, more specifically refreshable braille displays. It is expected that
+						a revision is needed to better account for the use case of embossing publications on paper.
 				</ul>
 			</section>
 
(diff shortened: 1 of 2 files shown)


bertfrees:
There is at least one thing that I expect will need to be changed, namely that the reading system requirements (https://github.com/daisy/ebraille/pull/192/commits/7620d96888b3bab94f870abd018be5b974dd9f3e) should probably be moved to a separate section.

clapierre (approved):

bertfrees:
My last commit provides the structure I have in mind for the CSS best practices. It is as follows:

- Media queries
- Spacing
- Indentation
- Alignment, positioning and spatial arrangement
- Text spacing
- White space handling
- Wrapping and breaking text
- Page breaking and column layout
- Borders and lines
- Rendering of tables
- Automatic numbering
- Floating content

[Preview](https://raw.githack.com/daisy/ebraille/css/best-practices/styling/index.html)

Next I will work on arranging the existing examples according to this structure. We also need more examples. Perhaps I should indicate in the document for which requirements we're still looking for examples, and then ask the group at the next meeting.

wfree-aph (commented):
Would only suggest minor changes.

Line 876: add word "as" so it reads "regarded as unreliable"
Line 876-877- remove "be able to" as it is redundant

index.html
```diff
@@ -63,6 +63,30 @@
 						"title": "Digital Publishing WAI-ARIA Module",
 						"href": "https://www.w3.org/TR/dpub-aria/",
 						"publisher": "W3C"
+					},
+					"css-page": {
+						"aliasOf": "css-page-3-20230914"
```
mattgarrish, line 68: It's best not to alias dated drafts like this as we'll be stuck with the dated version as they get updated. Is there a reason you can't use "css-page-3", for example, as those references always point to the latest published version.

mattgarrish (approved):
I've only read over the spec edits so far, but looks fine to me from a technological perspective. One comment about the aliasing of specifications, but otherwise just some minor edits I can make later (e.g., best to avoid "author" as it's caused nothing but confusion in epub with the actual content author and while I don't expect the same problems here it'd be good to keep our terminology aligned).

wfree-aph (approved):
Thanks for making those two minor changes! Looks great to me!

bertfrees:
I added a request for more examples in the CSS best practices, and two other harmless commits just now. I will merge the PR before today's meeting.

Rearrangement of best practices document will come in a next PR.

Issue #51, opened 2023-01-23
Possibility for optional hyphens and no-break spaces


jrbowden:
If a braille file, prepared for a particular line length, is opened on a reading system with a different line length, to optimise the braille, it would be useful that long words can be hyphenated appropriately, or that text that should not be split, remains together.

Examples:
The word "responsibilities" (to give but one) is a long word with no contractions in English braille. Due to word wrapping, without hyphenation, it can cause a bad line break.

Though not always considered essential, it should be possible to insert hyphenation points into such long words to allow good spllitting if needed.

Along similar lines, email and web addresses can be extremely long "words" and introducing hyphenation points can prevent splitting at inappropriate points. "Inappropriate" would be just to cram as much onto the available braille line and wrap regardless (worst case scenario).

On the other hand, in other situations it is considered good practice to prevent a line break between certain words - for example "chapter 3", "25 m" or "the number 2 000 000". Using no-break spaces would prevent inappropriate line breaking in these cases.

Of course, unless "advanced" searches are used, an ordinary text search should skip over optional hyphens and should treat no-break spaces as ordinary spaces.

Proposal:
Concepts such as an optional hyphenation point, no-break space and other typographical helpers should be available in the braille file.
Optional hyphens should also include what braille sign should be shown if the hyphen were inserted (it varies according to braille code).

JakeKyle:
How could cases be dealt with where the split hyphenated word is transcribed differently from the unhyphenated. I think @bertfrees is referring to this in [#4](https://github.com/daisy/ebraille/issues/4#issuecomment-1404647521).

There are several examples in UEB 2013 rule book section 10.13 "Word Division". One such is "inconvenient" in grade 2 UEB:

unhyphenated: ⠔⠉⠕⠝⠧⠢⠊⠢⠞

hyphenated: ⠊⠝⠤ ⠉⠕⠝⠧⠢⠊⠢⠞

How could this word with be represented in ebraille so that if it needed to be hyphenated it would obey the rules? I suppose all the alternative renderings would need to be part of the file?

Apologies if this has been dealt with elsewhere but I can't find it.

bertfrees:
Another example I recently learned is "comprendre" in French. It is normally contracted to "⠤⠖⠢⠹⠑", but when the line is broken after "com", it becomes:

    ................  ⠉⠺⠤
    ⠖⠢⠹⠑

[...]

PR #193, opened 2024-06-24
Add text restriction on xhtml character data

mattgarrish:
This pull request recommends using only characters in the unicode braille patterns set as well as html whitespace for renderable text (as best as I could define that). I don't know that we want to make it a stronger restriction as there may be some cases where non-braille characters are needed. I mentioned mathml in the note, as I expect that would be one case (but let me know if I'm wrong).

It also clarifies that this recommendation only applies to the default rendition if adding multiple renditions.

* [Preview](https://raw.githack.com/daisy/ebraille/spec/html-char/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/html-char/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -755,12 +755,13 @@
 						braille cells.</p>
 
 					<p>For this reason, the text content of [=eBraille content documents=] SHOULD consist only of the
-						following characters:</p>
+						following [[unicode]] characters:</p>
 
 					<ul>
-						<li><a href="https://www.unicode.org/charts/nameslist/c_2800.html">Braille Patterns</a>
-							[[unicode]];</li>
-						<li>[=ASCII whitespace=] [[infra]]</li>
+						<li><a href="https://www.unicode.org/charts/nameslist/c_2800.html">Braille Patterns</a> (U+2800
+							&#8230; U+28FF) ;</li>
+						<li>[=ASCII whitespace=] as defined in [[infra]] (U+0009, U+000A, U+000C, U+000D, and
+							U+0020)</li>
 						<li>NO-BREAK SPACE (U+00A0)</li>
 						<li>SOFT HYPHEN (U+00AD)</li>
 					</ul>
--- a/index.html
+++ b/index.html
@@ -759,9 +759,15 @@
 
 					<ul>
 						<li><a href="https://www.unicode.org/charts/nameslist/c_2800.html">Braille Patterns</a> (U+2800
-							&#8230; U+28FF) ;</li>
-						<li>[=ASCII whitespace=] as defined in [[infra]] (U+0009, U+000A, U+000C, U+000D, and
-							U+0020)</li>
+							… U+28FF) ;</li>
+						<li>[=ASCII whitespace=] [[infra]]: <ul>
+								<li>TAB (U+0009)</li>
+								<li>LF (U+000A)</li>
+								<li>FF (U+000C)</li>
+								<li>CR (U+000D)</li>
+								<li>SPACE (U+0020)</li>
+							</ul>
+						</li>
 						<li>NO-BREAK SPACE (U+00A0)</li>
 						<li>SOFT HYPHEN (U+00AD)</li>
 					</ul>
--- a/index.html
+++ b/index.html
@@ -785,8 +785,8 @@
 						attribute in <a href="#page-list">the page list</a>).</p>
 
 					<div class="note">
-						<p>For some technologies, such as [[mathml]], it will not be possible to restrict the text to
-							braille characters.</p>
+						<p>For some technologies that can be embedded in eBraille content documents, such as [[mathml]],
+							it will not be possible to restrict the text to braille characters.</p>
 					</div>
 				</section>
 
--- a/index.html
+++ b/index.html
@@ -511,7 +511,9 @@
 						rendition</a> [[epub-multi-rend-11]] (i.e., one listed first in the <a
 						data-cite="epub-33#sec-container-metainf-container.xml"><code>container.xml</code> file</a>
 					[[epub-33]]) is subject to the <a href="#fileset-structure">file naming and location
-						requirements</a> for the package document and primary entry page.</p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -738,18 +740,39 @@ <h3>Introduction</h3>
...
+						encodings that require a translation to 6- or 8-dot braille characters.</p>
+
+					<p>For this reason, the text content of [=eBraille content documents=] SHOULD consist only of
+						characters in the <a href="https://www.unicode.org/charts/nameslist/c_2800.html">Braille
+							Patterns</a> [[unicode]] as well as [=ASCII whitespace=] [[infra]].</p>
+
+					<div class="note">
+						<p>This restriction only applies to the renderable text of an XHTML document as well as any
+							attributes that could be rendered depending on the display capabilities of the user's device
+							(e.g., the [^img/alt^] attribute [[html]] for images that cannot be displayed).</p>
```
bertfrees, line 764: Another important one that will be used in eBraille is `abbr` on `th` elements. And perhaps we need to remind that in the [page list](https://raw.githack.com/daisy/ebraille/spec/html-char/index.html#example-page-list-with-front-and-body-matter) we explicitly require that the `title` attribute of entries is NOT in braille.

HTML defines [translatable attributes](https://html.spec.whatwg.org/multipage/dom.html#translatable-attributes), but I guess we can't refer to that?

index.html
```diff
@@ -738,18 +740,39 @@ <h3>Introduction</h3>
...
 
-				<section id="html-no-support">
-					<h3>Unsupported features</h3>
+					<p>[=eBraille content documents=] are required to provide their text content in 6- or 8-dot braille
+						characters. eBraille is not meant as a delivery format for ASCII braille or similar braille
+						encodings that require a translation to 6- or 8-dot braille characters.</p>
+
+					<p>For this reason, the text content of [=eBraille content documents=] SHOULD consist only of
+						characters in the <a href="https://www.unicode.org/charts/nameslist/c_2800.html">Braille
+							Patterns</a> [[unicode]] as well as [=ASCII whitespace=] [[infra]].</p>
```
bertfrees, line 759: I don't think ASCII white space will suffice. There is the [requirement of the soft hyphen](https://github.com/daisy/ebraille/issues/51). There might also be a need for zero-width space (to create soft wrap opportunities), and no-break space (although we may also use U+2800 for that).

index.html
```diff
@@ -738,18 +740,39 @@ <h3>Introduction</h3>
...
-					<li>MUST NOT contain features specified in <a href="#html-no-support"></a>.</li>
-				</ul>
+				<section id="xhtml-char-encoding">
+					<h4>Character encoding</h4>
 
-				<section id="html-no-support">
-					<h3>Unsupported features</h3>
+					<p>[=eBraille content documents=] are required to provide their text content in 6- or 8-dot braille
+						characters. eBraille is not meant as a delivery format for ASCII braille or similar braille
+						encodings that require a translation to 6- or 8-dot braille characters.</p>
```
bertfrees, line 755: If you put it this way, it seems ASCII braille is something in between that requires some processing that Unicode braille does not require. But to me Unicode braille and various ASCII braille variants are just different encodings of braille. They're on the same level. (But of course we prefer to follow the Unicode standard, and from the Unicode perspective, ASCII braille is odd).

wfree-aph (commented):
I agree with Bert's comment about soft-hyphen, no break space, and zero-width space being needed in addition to white space. I wasn't thinking about those aspects when we first discussed this idea.

index.html
mattgarrish: Ya, good point. I didn't mean it as translation to the unicode characters but the reading system has to know how to map the ascii braille to the correct display. I'll see if I can improve on it.

index.html
mattgarrish: > HTML defines [translatable attributes](https://html.spec.whatwg.org/multipage/dom.html#translatable-attributes), but I guess we can't refer to that?

No, that's not quite the same thing. It would allow braille in lang and style attributes, for example. They don't define attributes in the same kinds of classes as elements. Maybe it would make sense to provide a more complete list but without making the list exclusive.

bertfrees: Note that it does make sense in style attributes, although we probably shouldn't go there. To illustrate: in DAISY Pipeline, when it transcribes an HTML document to an HTML document with braille as text content, it goes into the style attributes and translates the "content" properties.

mattgarrish: Would title on abbr be another case, or how does braille software handle expansions? I always think of an AT possibly announcing the expansion, so needing tts-ready text, but does braille software provide the expansion in braille?

bertfrees: Good question. I would seem logical but I really don't know. I don't know if AT even uses title on abbr for TTS.

I can also imagine it would in some cases create funny situations when an expansion is provided but the expansion is actually abbreviated because abbreviations/contractions are baked into the braille code. Perhaps it would make more sense to provide expansions in uncontracted braille.

index.html
bertfrees: The wbr tag is probably better, because more explicit.

index.html
mattgarrish: They can use the title but typically it's a configuration option the user has to enable, like in jaws. I'd assume a braille device would provide the same option to use the expansions in place of the default text. It's probably worth mentioning it, as I don't expect we'd want plain text used.

But I also get confused about where voicing comes into play as I'm too used to making mainstream publications more accessible. I put aria-label on the nav elements without braille text in the navigation document examples thinking an AT would use those for announcing landmarks. Is that a realistic use case for braille? In other words, we wouldn't want to impose this restriction on aria-label even though it's an attribute whose text could sometimes be rendered.

bertfrees (approved):

PR #204, opened 2024-06-26
Packaging and reading system requirements

mattgarrish:
This pull request expands the packaging section to require ebraille publications be zipped in an ocf zip container when distributed. I also recommend against using the font obfuscation algorithm to embed fonts since we're recommending in the css section that epub creators not set font properties. (We could make this a must not, but then it seems like there should be a stronger ban on font-family changes.)

In trying to add the reading system requirements, I noticed that the reading system section was getting a bit big and hard to read. To try and make it more readable, I've made the following changes:

- there's now a general requirements section that covers how we diverge from epub
- there's a new unsupported features section that covers the features we're telling epub creators not to use but that aren't critical to security or privacy. As I expect ebraille reading systems will likely leverage what has already been done for epub, I don't know that there's a lot of value in forcing developers to spend time removing support that might exist. We don't want them to spend time implementing the features, but that's covered in the concluding paragraph about not adding support.
- support for packaged and unpackaged publications is covered in the publication types section
- the reading system requirements on 1ch and 1em are combined with not supporting font obfuscation in a section on fonts
- the requirement to support PDF is in a section on core media types

As always, feedback welcome on these changes.

* [Preview](https://raw.githack.com/daisy/ebraille/spec/packaging/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/packaging/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -880,10 +880,6 @@
 								font used by the reading device, using absolute lengths is regarded as unreliable. Using
 								font-relative lengths in style sheets is also needed in order to reliably use relative
 								widths and heights in <a href="#media-queries">media queries</a>.</p>
-							<p>Reading devices MUST guarantee that <code>1ch</code> corresponds with the cell-to-cell
-								distance (i.e. the distance from center to center of corresponding dots in adjacent
-								cells), while <code>1em</code> should correspond with the line-to-line distance (i.e.
-								the distance from center to center of corresponding dots from one line to the next).</p>
 						</li>
 					</ul>
 
@@ -1425,7 +1421,7 @@
 				in an EPUB container, as defined in <a data-cite="epub-33#sec-container-zip">OCF ZIP Container</a>
 				[[epub-33]].</p>
 
-			<p>Packaged eBraille publications are identified by the media type identifier
+			<p>Packaged eBraille publications are also identified by the media type identifier
 					"<code>application/epub+zip</code>" [[epub-33]]. The requirements for specifying the identifier in
 				the <code>mimetype</code> file are defined in <a data-cite="epub-33#sec-zip-container-mime">OCF ZIP
 					container media type identification</a>.</p>
@@ -1433,9 +1429,9 @@
 			<p>Unlike EPUB, which does not require a file extension, a packaged eBraille file set MUST use the extension
 					<code>.ebrl</code>.</p>
 
-			<p>As eBraille does not support <a href="#">CSS font properties</a>, eBraille creators MUST NOT use EPUB's
-					<a data-cite="epub-33#sec-font-obfuscation">font obfuscation algorithm</a> [[epub-33]] to embed
-				fonts.</p>
+			<p>As the use of <a href="#css-req">CSS font properties</a> is not recommended, eBraille creators SHOULD NOT
+				use EPUB's <a data-cite="epub-33#sec-font-obfuscation">font obfuscation algorithm</a> [[epub-33]] to
+				embed fonts.</p>
 		</section>
 		<section id="ebrl-security-privacy">
 			<h2>Security and privacy</h2>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1421,7 +1417,21 @@ <h2>Accessibility</h2>
...
+			<p>When [=eBraille publications=] are distributed to end users as a single file set, they MUST be packaged
+				in an EPUB container, as defined in <a data-cite="epub-33#sec-container-zip">OCF ZIP Container</a>
+				[[epub-33]].</p>
+
+			<p>Packaged eBraille publications are also identified by the media type identifier
+					"<code>application/epub+zip</code>" [[epub-33]]. The requirements for specifying the identifier in
+				the <code>mimetype</code> file are defined in <a data-cite="epub-33#sec-zip-container-mime">OCF ZIP
+					container media type identification</a>.</p>
+
+			<p>Unlike EPUB, which does not require a file extension, a packaged eBraille file set MUST use the extension
```
clapierre, line 1429: I didn't know that EPUB's didn't required .epub.  Should we at least say although the recognized extension is .epub?
just seems like an obvious question folks will have when reading this statement.

index.html
```diff
@@ -1516,55 +1526,107 @@ <h3>Introduction</h3>
...
+						<li><a data-cite="epub-rs-33#sec-epub-rs-conf-backward">backward compatibility</a> or <a
+								data-cite="epub-rs-33#sec-epub-rs-conf-forward">forward compatibility</a> with other
+							EPUB versions.</li>
+					</ul>
 
-				<p>In addition, eBraille reading systems MUST support PDF for tactile graphics. If an eBraille reading
-					system is not capable of rendering PDFs embedded in [=eBraille content documents=], it MUST provide
-					a mechanism that allows the user to open the graphics in an application that supports PDF
-					rendering.</p>
+					<p>eBraille reading systems not built on EPUB rendering engines MUST NOT add support for these
```
clapierre, line 1579: Did you really mean "not build on EPUB rendering engines"?  seems if they were build on these EPUB rendering engines they must remove that support.

index.html
mattgarrish: I don't think we need to bring in epub here. The requirement is a .ebrl extension, full stop. I've added a note to explain that if the file has a .epub extension it won't be an ebraille file anymore but might open in an epub reading system with some loss of functionality. That might be more useful to help people understand how the two relate.

index.html
mattgarrish: I've changed the name to "Non-critical features", as "Unsupported features" is too broad. There are critical features that must not be supported in the first section. I've also tried to explain this better in the first paragraph.

clapierre (approved):

wfree-aph (approved):
Looks good! Thanks Matt!

index.html
```diff
@@ -1516,55 +1532,109 @@ <h3>Introduction</h3>
...
+								resources</a> [[epub-rs-33]].</li>
+						<li>It MUST NOT resolve [=path-absolute-URL strings=] [[url]].</li>
+						<li>As remote resources, forms, and scripting are not supported, it MUST NOT allow [=eBraille
+							publications=] to have <a data-cite="epub-rs-33#sec-epub-rs-network-access">network
+								access</a> [[epub-rs-33]]. (Note that barring network access should not prevent users
+							from following <a data-cite="epub-rs-33#sec-epub-rs-external-links">external hyperlinks</a>
+							[[epub-rs-33]] in a publication.)</li>
+						<li>It MUST only support [=eBraille content documents=] in the <code>spine</code>. The reading
+							system MAY reject eBraille publications with other formats in the spine or ignore their
+							entries when rendering the publication.</li>
```
egli, line 1555: Maybe this is a lack of understanding on my part, but how can you have multiple renditions in an eBraille, e.g. braille and audio and at the same time the spine only contains braille?

Shouldn't we be less restrictive with the spine in light of multiple renditions?

mattgarrish: My understanding of multiple renditions is that they're wanted for providing text for tts. If you're trying to put audio in the spine, then we'd have to bring back manifest fallbacks as there's no ability to do this without including a fallback to a text document (this is true for epub, as well). If you want media overlays in an alternative rendition, you still need to have text documents in the spine to synchronize with the audio.

There's an exception in the multiple renditions section that doesn't require additional renditions to meet the braille character encoding requirements, so that case is supported. The xhtml still has to conform to the ebraille content document definition, but that shouldn't be problematic if the text structure is the same as the braille.

egli: After having talked to @ManfredMuchenberger I understand now. All fine.

Issue #194, opened 2024-06-25
Rename brl:dotNumber to brl:cellType


egli:
Makes the intention behind the field more clear

mattgarrish:
Closing this issue as the name change was included in #209

mattgarrish:
Reopening this because of the proposed change of prefix from brl: to a11y: in #216 

The [question](https://github.com/daisy/ebraille/pull/216#pullrequestreview-2156979917) is whether the property still makes sense without a braille sounding prefix to it.

wfree-aph:
@mattgarrish It seems like it doesn't really make sense without the braille association in the prefix. Could it be changed to a11y:braillecelltype?

mattgarrish:
Yes. The only thing I'd suggest differently is to keep the camel case naming we use for the other properties, so make it a11y:brailleCellType

PR #209, opened 2024-06-28
Integrate metadata guide

mattgarrish:
This pull request moves the metadata requirements into the main specification.

I'm opening this pull request a bit early to start getting feedback. I've tried to fill in some of the sections as best I can with some more details from how epub defines/handles them, but I may have some things wrong (especially whether some properties can be repeated). We also discussed some additional changes in the last meeting that I haven't tried to go back and integrate yet. Let me know whatever you spot that needs fixing or needs more detail.

* [Preview](https://raw.githack.com/daisy/ebraille/spec/metadata/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/metadata/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -802,8 +802,8 @@
 						<p>The REQUIRED <code>brl:code</code> property identifies the name of the braille code an
 							[=eBraille publication=] has been formatted in conformance with.</p>
 
-						<p>The value SHOULD include any code-specific information needed to fully its formatting (e.g.,
-							"contracted", "uncontracted", "bar by bar" etc.).</p>
+						<p>The value SHOULD include all relevant code-specific information (e.g., "contracted",
+							"uncontracted", "bar by bar" etc.).</p>
 
 						<div class="ednote">
 							<p>A list of recommended braille code terminology will be provided with a future update to
--- a/index.html
+++ b/index.html
@@ -619,6 +619,720 @@
    &#8230;
 &lt;/package></code></pre>
 				</aside>
+
+				<section>
+					<h4>Dublin Core Required Metadata</h4>
+
+					<section id="dc:creator">
+						<h4>dc:creator</h4>
+
+						<p>The content of dc:creator shall be the name(s) of the primary author, editor, etc. (when
+							appropriate). Follow regional conventions for name layout. Note: For multiple creators, this
+							element shall be repeated as needed. </p>
+
+						<dl class="elemdef">
+							<dt>Allowed values:</dt>
+							<dd>
+								<ul class="nomark">
+									<li>Any string</li>
+								</ul>
+							</dd>
+
+							<dt>Examples</dt>
+							<dd>
+								<aside class="example"
+									title="A single author with an English name arraigned with the last name appearing first and then followed by a comma and then the first name">
+									<pre><code class="html">&lt;metadata ...>
+	&lt;dc:creator>
+		Twain, Mark
+	&lt;/dc:creator>
+	...
+&lt;/metadata></code></pre>
+								</aside>
+								<aside class="example"
+									title="A single author with a Japanese name arraigned with first name appearing first and then followed by last name with no comma">
+									<pre><code class="html">&lt;metadata ...>
+	&lt;dc:creator>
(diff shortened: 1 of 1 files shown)


wfree-aph (commented):
Line 713 shows >well-formed. I wasn't sure if that greater than symbol was necessary or a typo.

Only other thing is that at line 1090 with Optional properties. I think we decided as a group to just have required and recommended with the idea being that we just wouldn't define "optional" and people could add what they wanted as long as they followed the required properties.

bertfrees (approved):
There is something wrong in the sentence "The value SHOULD include any code-specific information needed to fully its formatting".

Shouldn't dc:language always include the "Brai" subtag, since the content is, per definition, in braille script? Perhaps it should not be required, since it can be assumed. But regardless, wouldn't it be more accurate if the examples in the spec were changed?

Then there's one other thing. I know this is probably not the right place and time to discuss it, but I only found out about the brl:dotNumber field now. Can someone fill me in on why exactly it was requested? This is typically information that I would expect in the brl:code field. I could imagine an additional field could be useful for a user agent to quickly test whether a publication may be rendered (if it depends on whether 8-dot cells are present or not), but then I don't get why it should be allowed to have "6, 8" or "8, 6". Sorry if I'm bringing up things that have been discussed before. If someone can please just briefly fill me in. If I have more questions I will take it up in a separate issue :-)

mattgarrish:
> There is something wrong in the sentence "The value SHOULD include any code-specific information needed to fully its formatting".

Ya, something weird happened there. I've fixed that to:

> The value SHOULD include all relevant code-specific information (e.g., "contracted", "uncontracted", "bar by bar" etc.).

mattgarrish:
> Perhaps it should not be required, since it can be assumed.

Ya, I don't know about requiring it when we have multiple renditions. The text rendition wouldn't be braille.

I'm okay with adding -Brai to the examples, though, since we're focused on the default braille rendition.

mattgarrish:
> Line 713 shows >well-formed. I wasn't sure if that greater than symbol was necessary or a typo.

Necessary. It's just the line formatting from the editor I use. It broke the closing angle bracket for the tag onto a new line because of the length of the URL on the previous plus the word after the tag wouldn't fit any other way.

mattgarrish:
> the idea being that we just wouldn't define "optional" and people could add what they wanted as long as they followed the required properties

Yes, and I don't plan to add any optional property definitions there. But I think we should still explain that all the DC elements are available because we are using epub and you can add prefixes for other vocabularies. Otherwise, I think people not familiar with the epub package document might get confused that we don't allow anything but the required and recommended properties.

mattgarrish:
> I could imagine an additional field could be useful for a user agent to quickly test whether a publication may be rendered ... but then I don't get why it should be allowed to have "6, 8" or "8, 6"

I'm probably not the right person to answer this, but wouldn't it allow the reading system to attempt to render the publication (if it's predominantly in a cell type it supports) rather than rejecting it outright because it can't render any of it?

In any case, I've made the name change to cellType that was request in the last call and logged in #194 .

egli:
> Can someone fill me in on why exactly it was requested? This is typically information that I would expect in the brl:code field. I could imagine an additional field could be useful for a user agent to quickly test whether a publication may be rendered (if it depends on whether 8-dot cells are present or not), but then I don't get why it should be allowed to have "6, 8" or "8, 6". Sorry if I'm bringing up things that have been discussed before. If someone can please just briefly fill me in. If I have more questions I will take it up in a separate issue :-)

I think the motivation was as you mention so that a user agent can quickly test whether it can support a publication.

As far as I understand the reason for allowing both values is so you can publication that contain both kinds of braille cells.

mattgarrish:
I'm going to merge this pull request before it becomes any bigger and then we can take up individual issues in separate pull requests.

PR #216, opened 2024-07-03
Switch from brl: to a11y: prefix for metadata properties

### LINKED ISSUE #215: Switch brl: prefix to a11y:
Although it's possible to add a new reserved prefix, it comes with a cost of full compatibility with EPUB 3 unless a prefix is always declared or we try to get the epub spec amended to add a new reserved prefix (which could be difficult as it's probably a feature request and epub is in maintenance mode).

To avoid these issues, it's probably better to define the needed vocabulary terms in the existing a11y: prefix namespace. That will give us full compatibility without the prefix declaration headaches.

It has the side benefit of making the terms we define more generally usable for future accessible formats. Using a brl: prefix suggests they are only usable for braille, while some like an a11y:contributor property would be useful no matter the format.

mattgarrish:
In addition to renaming the properties and updating the section on referencing them, I've also removed all the problematic compatibility notes about brl: not being an epub reserved prefix.

Note that this doesn't change the naming of any existing properties or add any new ones. We should tackle those as separate issues.

Fixes #215 

* [Preview](https://raw.githack.com/daisy/ebraille/metadata/prefix/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/metadata/prefix/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -600,9 +600,9 @@
 							<dd>
 								<p>Dublin Core defines 15 core elements in the <code>/elements/1.1/</code> namespace
 									[[dcterms]] that can be used anywhere in the package document. Elements not in the
-										<a href="#meta-req-prop">required</a> or <a href="#meta-rec-prop"
-										>recommended</a> metadata sections are valid to use &#8212; these are considered
-										<a href="#meta-additional">optional elements</a>.</p>
+										<a href="#meta-req">required</a> or <a href="#meta-rec">recommended</a> metadata
+									sections are valid to use &#8212; these are considered <a href="#meta-additional"
+										>optional elements</a>.</p>
 							</dd>
 
 							<dt>The <code>meta</code> element</dt>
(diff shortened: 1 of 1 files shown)


bertfrees (approved):

clapierre (commented):
I think this is fine to start, but some items may need further refinement where just having a11y:xyz won't make sense such as a11y:cellType here I would think a11y:brailleCellType.

mattgarrish:
Ya, it's easy to get into bikeshedding names with metadata which is why it'd be better to open separate issues for each name that could be changed. I'll reopen #209 as that was already about picking a better name and led to cellType.

GeorgeKerscher:
I think this looks right, swaping in the A11y prefix.

mattgarrish:
I found epubcheck doesn't like additional properties while creating the sample, but not sure why. I don't think it verifies any other vocabularies to make sure properties exist.

In any case, that's a minor technicality as we need an updated validator regardless of what we define.

PR #222, opened 2024-07-08
Updated abstract

mattgarrish:
I noticed that our abstract wasn't really saying anything specific about the format, as I expect it was a placeholder from when we created the document. This adds another paragraph to it so it now reads:

> This specification defines eBraille, a digital reading format for braille publications.
> 
> The eBraille file set is based on Open Web Technologies, such as XHTML and CSS, and provides the ability to encode braille in semantically enhanced markup, allowing it to adapt to the different capabilities of braille reading devices. The file set is designed for both packaged distribution to end users and deployment to the web for online and downloadable reading.

Let me know if there's anything else we should highlight in the abstract, or that doesn't need saying in the above.

* [Preview](https://raw.githack.com/daisy/ebraille/spec/abstract/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/abstract/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -100,8 +100,9 @@
 			<p>This specification defines eBraille, a digital reading format for braille publications.</p>
 
 			<p>The eBraille file set is based on Open Web Technologies, such as XHTML and CSS, and provides the ability
-				to encode braille in semantically enhanced markup. It is designed for both packaged distribution to end
-				users and deployment to the web for online and downloadable reading.</p>
+				to encode braille in semantically enhanced markup, allowing it to adapt to the different capabilities of
+				braille reading devices. The file set is designed for both packaged distribution to end users and
+				deployment to the web for online and downloadable reading.</p>
 		</section>
 		<section id="sotd">
 			<div data-include="common/status.html" data-include-replace="true"></div>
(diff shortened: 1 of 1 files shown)


clapierre (approved):
Looks good, did we want to mention anything about the packaging format of EBRF is based on EPUB?

mattgarrish:
I don't know if that's important for the abstract since we don't really expect the publications to be read in epub reading systems, but I don't have a strong opinion on it. The epub connection gets fully covered in the overview and "relationship to epub 3" sections.

If we wanted to, I guess we could rewrite the start of the second paragraph to something like:

> eBraille uses an EPUB 3-compatible file based on Open Web Technologies ...

bertfrees (approved):

mattgarrish:
I did some other minor tweaks so the second paragraph now reads:

> eBraille uses an EPUB 3-compatible file set based on Open Web Platform — using technologies such as XHTML and CSS — to encode braille in semantically enhanced markup and allow it to adapt to the different capabilities of braille reading devices. The file set is designed for both packaged distribution to end users and deployment to the web for online and downloadable reading.

clapierre (approved):
Much better looks great.

PR #243, opened 2024-08-08
Reorder creator names and add file-as example

### LINKED ISSUE #218: Creator name formatting
I noticed that the dc:creator examples have the last name first followed by a comma. For example:

```
<dc:creator id="aut01">
   Brontë, Charlotte
</dc:creator>

<dc:creator id="aut02">
   Melville, Herman
</dc:creator>
```

If a reading system merges these using commas in the bookshelf, however, the user would likely be presented with:

>    Brontë, Charlotte, Melville, Herman

This could get confusing to read. Is it necessary to format the names this way, or can we use the file-as property to provide sorting information? For example:

```
<dc:creator id="aut01">
   Charlotte Brontë
</dc:creator>
<meta property="file-as" refines="#aut01">Brontë, Charlotte</meta>
```

mattgarrish:
I've also added a change log section to start listing off the substantive changes between drafts.

Fixes #218

### DIFF
--- a/index.html
+++ b/index.html
@@ -2998,6 +2998,29 @@
 				</section>
 			</section>
 		</section>
+		<section id="change-log" class="informative">
+			<h2>Change log</h2>
+
+			<div class="note">
+				<p>This section identifies substantive changes made to the eBraille specification between drafts. For a
+					list of all changes, refer to the <a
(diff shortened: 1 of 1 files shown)


clapierre (approved):

wfree-aph (approved):
Thank you for correcting this issue! I was still unsure of the correct way to word everything and should have reached out for more information sooner.

PR #253, opened 2024-08-22
Remove "various strings," adds wbr and &shy; examples, and minor cleanup

### LINKED ISSUE #224: Add an example that uses `<wbr>` in tagging best practices
In https://github.com/daisy/ebraille/pull/193 we added the recommendation to indicate break opportunities within words using the `<wbr>` element. We should have an example in the tagging best practices document that shows how to use this element.

### LINKED ISSUE #227: Reserved class names without any restrictions make no sense
If a class name can be anything (any valid CSS identifier), what does it mean to "reserve" it? For example:

- For boxed text: on `div`: _various strings_

   The explanation in the text is:

   > Various strings can be used for boxes with specific text needs. Examples could include boxes with color. There is no way to define all possible strings that could be used in such a scenario. The writing software will need to define the limitations and allow the user to input the string, which is then used as the class to allow the CSS to input that string into the box line.

   This is correct, but shouldn't writing software allow the user to define new classes _anyway,_ for potentially _any_ element? It is good to include the examples of colored boxes, but it makes no sense to add "various strings" to the reserved class names because 1. what does it mean to "reserve various strings", and 2. we could add it everywhere.

wfree-aph:
Removed "various strings" from box reserved class values.

Replaced br with wbr in multiline heading example. Also added a note explaining the preference for wbr over br.

Closed a p tag. Added a comma and space where it was previously missing in the paragraph reserved class values.

Fixes #227
Fixes #224

***
<a href="https://pr-preview.s3.amazonaws.com/daisy/ebraille/pull/253.html" title="Last updated on Aug 29, 2024, 12:53 PM UTC (d4232c1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/daisy/ebraille/253/aa004f1...d4232c1.html" title="Last updated on Aug 29, 2024, 12:53 PM UTC (d4232c1)">Diff</a>

### DIFF
--- a/best-practices/tagging/index.html
+++ b/best-practices/tagging/index.html
@@ -107,7 +107,7 @@
 					<dt>Reserved <code>class</code> values:</dt>
 					<dd>
 						<ul class="nomark">
-							<li><code>box</code>, <code>nemeth</code>, various strings</li>
+							<li><code>box</code>, <code>nemeth</code></li>
 						</ul>
 					</dd>
 				
@@ -302,8 +302,9 @@
 				</aside>
 
 				<aside class="example" title="Multiline heading">
-					<pre><code class="html">&lt;h1>⠠⠉⠓⠁⠏⠞⠑⠗ ⠼⠛&lt;br/>
-⠠⠞⠓⠑ ⠠⠗⠑⠍⠁⠗⠅⠁⠃⠇⠑ ⠠⠞⠗⠥⠞⠓&lt;br/>
+					<p>Note the use of <code>&lt;wbr/></code> instead of <code>&lt;br/></code>. <code>&lt;wbr/></code>, being optional, will only be utilized if the display being used cannot accommodate the entire heading on one line.</p>
+					<pre><code class="html">&lt;h1>⠠⠉⠓⠁⠏⠞⠑⠗ ⠼⠛&lt;wbr/>
+⠠⠞⠓⠑ ⠠⠗⠑⠍⠁⠗⠅⠁⠃⠇⠑ ⠠⠞⠗⠥⠞⠓&lt;wbr/>
 ⠁⠃⠕⠥⠞ ⠠⠚⠕⠓⠝ ⠠⠎⠍⠊⠞⠓&lt;/h1></code></pre>
 				</aside>
 				<aside class="example" title="Running header, also called a running head or running heading when used for the title of the work being transcribed">
@@ -321,7 +322,7 @@
 					<dt>Element(s):</dt>
 					<dd>
 						<p data-cite="html">[^img^], [^a^]</p>
-						<p>Note that <code>img</code> is used for jpg, png, and svg, while <code>a</code> is used for pdf.
+						<p>Note that <code>img</code> is used for jpg, png, and svg, while <code>a</code> is used for pdf.</p>
 					</dd>
 
 					<dt>Reserved <code>class</code> values:</dt>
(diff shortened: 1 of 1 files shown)


clapierre (approved):
I didn't know about `<wbr>` thats cool

wfree-aph:
@bertfrees Is it okay to use `wbr` for an example and then add a separate example that demonstrates `text-wrap: balance`? Or do we only need one or the other?

bertfrees:
> `<wbr/>`, being optional, will only be utilized if the display being used cannot accommodate the entire heading on one line.</p>

I don't think it works like that. `<wbr/>` is for creating line break opportunities in places where a line would otherwise not be broken. E.g. it is typically used inside words.

>  Is it okay to use `wbr` for an example and then add a separate example that demonstrates `text-wrap: balance`? Or do we only need one or the other?

We could have the following examples:

- one for `br`, but it's good to mention here that by using this element you will hard-code the formatting to a certain line length, so we should add that it is to be avoided unless the publication is made for that specific line length. To emphasize the issue, we could perhaps come up with an example that uses `br` "the proper way", using a class and a media query:

  ```html
  <p>
    foo<br class="for-width-50"/> bar
  </p>
  ```
  ```css
  br[class^=for-width] {
    display: none;
  }
  @media (width > 49ch) and (width < 51ch) {
    br.for-width-50 {
      display: inline;
    }
  }
  ```

- one for `text-wrap: balance`: it won't be very exciting because it will not contain any special formatting, but we could perhaps link to the corresponding example in the styling best practices 
- one for `wbr` ([#224](https://github.com/daisy/ebraille/issues/224)), but that's an entirely different issue

bertfrees (changes requested):
See my comment about the wrong use of `wbr`.

wfree-aph:
@bertfrees Would you mind reviewing these changes? I think my latest commits fixes my mistakes from earlier while also addressing the use of `wbr`.

bertfrees (approved):
Looks good.

best-practices/tagging/index.html
```diff
@@ -300,8 +300,8 @@ <h4>Headings</h4>
 				<aside class="example" title="Simple heading">
 					<pre><code class="html">&lt;h1>⠠⠊⠝⠞⠗⠕⠙⠥⠉⠞⠊⠕⠝&lt;/h1></code></pre>
 				</aside>
-
 				<aside class="example" title="Multiline heading">
+					<p>Note that while <code>&lt;br/></code> can be used to force multiple lines, doing so will hardcode those breaks. It is better to identify the heading with either a tag or class and use CSS to center the braille &lpar;<code>text-wrap: balance</code>&rpar;. See <a href=https://daisy.github.io/ebraille/best-practices/styling/#line-breaking>Styling Best Practices, section 2.7</a>.</p>
```
bertfrees, line 304: I'm not sure what the best practice is when referencing other sections. E.g. should we include the number (because it could change)?

mattgarrish: No, don't use the section number. You can use just the section title, as they're less likely to change, but if you can write a descriptive text that doesn't rely on either that's even better (e.g., "for more information, refer to the the section on ... [brl-style]").

wfree-aph:
Removed that reference and now will merge. Thanks for the help cleaning it up!

bertfrees:
>   ```html
>   <p>
>     foo<br class="for-width-50"/> bar
>   </p>
>   ```
>   ```css
>   br[class^=for-width] {
>     display: none;
>   }
>   @media (width > 49ch) and (width < 51ch) {
>     br.for-width-50 {
>       display: inline;
>     }
>   }
>   ```

What do you guys think? Would it be good to include an example like the one I suggested above, to show how, when the layout needs to be optimized for a certain line length, it can be done in a way that it does not affect the layout for other line lengths?

It is perhaps somewhat advanced, but on the other hand an important selling point of eBraille is that it is dynamic, so it is useful to make people aware that there are some things you need to do to ensure that it stays dynamic.

wfree-aph:
@bertfrees I can see the value in including that example. Probably would be important to specify the circumstances you would use this advanced version over a simpler method maybe.

PR #257, opened 2024-08-29
Add media type registration

mattgarrish:
This pull request adapts the OCF media type registration from epub 3.3 for ebraille. It defines application/ebrl+zip for ebraille files.

I removed the magic number part for the epub media type because it doesn't seem useful to check that for ebraille and we also are looser about accepting it so it may not even be where it should be. Otherwise, I've just adapted the text to call this the "ebraille implementation of the EPUB OCF zip container".

Let me know if there's anything that you think needs adjusting in this.

* [Preview](https://raw.githack.com/daisy/ebraille/spec/media-type/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/media-type/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -2997,29 +2997,6 @@
 				</section>
 			</section>
 		</section>
-		<section id="change-log" class="informative">
-			<h2>Change log</h2>
-
-			<div class="note">
-				<p>This section identifies substantive changes made to the eBraille specification between drafts. For a
-					list of all changes, refer to the <a
-						href="https://github.com/daisy/ebraille/issues?q=is%3Aissue+is%3Aclosed">issue tracker</a>.</p>
-			</div>
-
-			<section id="changes-latest">
-				<h3>Changes since the <a href="https://daisy.org/s/ebraille/1.0/FPWD-ebraille-20240725/">First Public
-						Working Draft</a></h3>
-
-				<ul>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2998,20 +2997,137 @@ <h4>tactileGraphics</h4>
...
+								href="https://www.iana.org/assignments/media-types/application/zip">
+								<code>application/zip</code></a> media type.</p>
+					</dd>
+
+					<dt>Security considerations:</dt>
+					<dd>
+						<p>All processors that read OCF ZIP container files should rigorously check the size and
+							validity of data retrieved.</p>
+						<p>In addition, because of the various content types that can be embedded in OCF ZIP container
+							files, <code>application/epub+zip</code> may describe content that poses security
```
clapierre, line 3051: this should be ebrl+zip right?

clapierre (approved):
Looks good other than the ebrl+zip change.

index.html
mattgarrish: Oops, good catch! Missed that one in the copy of the text.

PR #259, opened 2024-09-05
Add explanatory note about use of epub media type in packaging

mattgarrish:
This pull request adds a note to clarify why the mimetype file has to have the epub media type (epub packaging compatibility) while a different media type is defined for ebraille publications in the appendix.

* [Preview](https://raw.githack.com/daisy/ebraille/spec/mimetype-expl/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/mimetype-expl/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -2428,10 +2428,18 @@
 				[[epub-33]] are only required for a packaged eBraille file. The files MAY be omitted if an eBraille
 				publication is not packaged.</p>
 
-			<p>Packaged eBraille publications use the same media type identifier as EPUB:
-					<code>application/epub+zip</code> [[epub-33]]. The requirements for specifying this identifier in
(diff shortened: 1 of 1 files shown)


clapierre (approved):

clapierre:
So, does this mean there are two <mimetype>'s Matt?  Would an example somewhere help?

clapierre:
> What if I mention that in the last sentence: "This is the media type used, for example, in the Content-Type header when serving packaged eBraille publications over the web."?

I think I like that idea.

bertfrees (approved):

Issue #244, opened 2024-08-13
Do something about "magic numbers" in CSS examples


bertfrees:
See for example the [Simple box with blank lines](https://daisy.github.io/ebraille/best-practices/styling/#example-simple-box-with-blank-lines) example:

```css
.box {
   border-style: solid;
   border-width: .45rem 0;
   margin: 1.22rem 0 1.55rem 0;
   padding: .33rem 0 0 0;
}
```

Either these numbers need an explanation (how to obtain the numbers?), or we need to recommend another approach that doesn't need the numbers (e.g. we could further develop the "[border image](https://daisy.github.io/ebraille/best-practices/styling/#example-border-image)" approach).

bertfrees:
See https://github.com/daisy/ebraille/pull/261

PR #261, opened 2024-09-09
Add tool for generating border images

bertfrees:
The idea is to eventually make it available as a server-side tool hosted on daisy.org, but for now a polyfill is required to resolve the special border-image URLs.

[Preview](https://raw.githack.com/daisy/ebraille/border-image-generator/best-practices/styling/index.html?developer-mode=true#borders-lines)

Being able to easily generate the source image for creating a border of braille characters is very convenient because the way the best practices document currently recommends creating borders is terribly complex (see the issue [Do something about "magic numbers" in CSS examples](https://github.com/daisy/ebraille/issues/244)).

The one issue I can see with this it that [we currently don't allow remote resources](https://daisy.github.io/ebraille/#res-location).


clapierre (approved):

best-practices/styling/index.html
```diff
@@ -1057,6 +1057,30 @@ <h3>Borders and lines</h3>
 							‘text-indent’ property</a></li>
 						</ul>
 					</details>
+					<div class="note">
+						<p>The same result may be achieved using the “border-image” approach
```
mattgarrish, line 1061: I guess this won't work until the scripts are merged, right? All I get is an empty css example box in the preview and locally.

bertfrees: Yes, sorry, I tested it locally by replacing

```js
script.src = "https://daisy.github.io/ebraille/best-practices/styling/handleBorderImageURLs.js";
```

with

```js
script.src = new URL("handleBorderImageURLs.js", document.baseURI).href;
```

bertfrees: I was talking about the "Validate with alternative CSS" buttons. But now I noticed that the CSS is actually missing. Hum... I have no idea how that happened.

mattgarrish: Ya, I couldn't figure out if the example css was disappearing because some scripts were running on the file but not others. It might be easier to merge and then do a final check that everything is okay.

wfree-aph:
@bertfrees Apologies for my ignorance but I'm not understanding how this works. It seems like this feature would require someone using an eBraille file to have access to the internet. Is that true? It also seems like a feature that would make support on reading systems more complicated. I can understand why you're looking for a different approach than what we have without this feature but I'm just intimidated by the implications of what you're suggesting.

bertfrees:
> It seems like this feature would require someone using an eBraille file to have access to the internet.

Yes, the way it is currently presented in the examples, it absolutely requires internet access. But nevertheless I think it would be better to recommend using the border-image approach for borders. I can see two solutions:

- The author uses the DAISY service to generate the images and includes it in the eBraille file (manually, or the authoring tool could do it)
- The author creates an eBraille file that uses remote resources, and then a preprocessing tool automatically downloads the resources, puts them inside the file, and updates the remote references.

wfree-aph:
> > I can see two solutions
> 
> A third solution is to create a CSS extension. We could provide the JavaScript library for reading system implementors to use.

Would it be possible for us to discuss these solutions? A meeting with @bertfrees, @mattgarrish, and me would help me understand the implications here. Right now, I feel like I'm still nervous about the implications of each and if I understood the possibilities better, it would help me know what you're thinking.

bertfrees:
I'm available for a meeting.

bertfrees:
Changes requested:

- Don't recommend using the remote URLs. Instead, explain how to use the image generator tool to generate and download an image, store it inside the eBraille publication, and reference it from the CSS.
- Mention that the `::before`/`::after` approach may be used as an alternative to the border-image approach, as long as there are no left and right borders.
- For completeness sake, also still mention the `border-width`/`padding`/`margin` approach, but keep it for last.
- To get a good overview, divide the examples into three subsections.

bertfrees:
@wfree-aph I updated the document. Could you please read through it to check if you can spot any other issues?

best-practices/styling/index.html
bertfrees: This wasn't resolved yet. It seems like highlightjs is messing with the code blocks within `details` elements. I don't understand what is happening. For now my workaround is to not mark up these code blocks with `<pre><code>`. @mattgarrish Can you think of a way to fix this properly?

PR #262, opened 2024-09-10
Adding samples to best practices > styling > CSS Best Practice Examples

wfree-aph:
Included are BRF samples in folders relating to the current sections of the Styling Best Practices document. The section numbers may eventually be out of date but needed some way to organize them. Let me know if you'd rather I didn't drop these examples here. Wanted somewhere that we could reference them and also so others could potentially review them.


bertfrees:
Thanks! Fine to put them here, but I would maybe leave out the BRFs that were already added before but then removed later as examples were added to the HTML page (see commits [`e254ac4^..9a78378`](https://github.com/daisy/ebraille/compare/e254ac43f283a76a96c46b25a0ede2865674c506^..9a783785b17a11e92fe7fed7414f56ce07d963c4) and 9714b89).

wfree-aph:
@bertfrees I was thinking that having all of the sample files available could be useful as a record of where the examples came from. What do you think? Should I remove the ones you're referencing? I'm fine to do it but just wanted to mention this idea and see if it changed anything.

wfree-aph:
It was discussed on the editor's call that completed braille examples be identified by placing them in a folder. I will add that folder and call it "Added to the spec." Once those changes are in place, this PR can be reviewed again.

wfree-aph:
Having a bit of trouble sorting these files out, so I'm going to merge this branch and then start adding examples to the styling best practices. As I move examples to that document, I'll put them in the folder indicating they are in the spec. Once I'm done with that, I'll see if there are any left that need to be handled.

PR #269, opened 2024-09-23
Add sourceDate property

### LINKED ISSUE #168: brl:producedDate instead of brl:created
This would follow Daisy's metadata dtd:producedDate. This is a proposal for a name change only, not its contents.

### LINKED ISSUE #200: Add brl:sourceDate
According to the [Specifications for the Digital Talking Book](http://www.daisy.org/z3986/2005/Z3986-2005.html):

> dtb:sourceDate
> Date of publication of the resource (e.g., a print original, ebook, etc.) from which the DTB is derived.
> Occurrence: Optional - recommended. Not repeatable.

### LINKED ISSUE #221: sourcePublisher required or recommended
Looking at the required metadata, having sourcePublisher in the list requires that the ebraille publication come from another source. Is there no expectation that the format could be used to create braille-original works?

If there is that possibility, then sourcePublisher should only be recommended as it's the one piece of required metadata that isn't about describing the ebraille publication itself.

mattgarrish:
This pull request adds the property for expressing the date on which the source was published.

Should we also move this and sourcePublisher to the recommended properties list, as I asked in #221? Looking at the sourcePublisher definition, it says to add "N/A" if there isn't a source publisher, which is a strong case for not making it required.

Required metadata should always have a meaningful value, otherwise we keep having to propagate a workaround. It's never great to require language-specific values like "N/A", which I'll have to work into the date if we keep them required.

Fixes #221 
Fixes #200 
Fixes #168 

* [Preview](https://raw.githack.com/daisy/ebraille/spec/issue-200/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/issue-200/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1209,7 +1209,26 @@
 
 						<p>Repeat the property for each organization or individual.</p>
 					</section>
-
+					
+					<section id="a11y:sourceDate">
+						<h5>a11y:sourceDate</h5>
+						
+						<p>The REQUIRED <code>a11y:sourceDate</code> property identifies the date on which the source
+							work was published.</p>
+						
+						<p>The property is defined in a <a href="#meta-elem"><code>meta</code> tag</a> with its
+							<code>property</code> attribute set to <code>a11y:sourceDate</code>.</p>
+						
+						<p>It is RECOMMENDED that the date string conform to [[iso8601]], particularly the subset
+							expressed in W3C Date and Time Formats [[datetime]].</p>
+						
+						<aside class="example" title="Date of a source publication">
+							<pre>&lt;meta property="a11y:sourceDate">
+   2020-03-15
+&lt;/meta></pre>
+						</aside>
+					</section>
+					
 					<section id="a11y:sourcePublisher">
 						<h5>a11y:sourcePublisher</h5>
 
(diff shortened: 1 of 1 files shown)


clapierre (changes requested):
I am concerned about the prefix "a11y" as these are not accessibility metadata.
Is there something more appropriate?

mattgarrish:
The source publisher and source publication date are pretty unique to accessible format production and we're using a11y as a prefix for metadata that is broadly needed for accessible formats. I wouldn't try to read too much into a prefix.

We're also constrained by epub's reserved prefixes. If we mint something new, then the prefix has to be declared every time until we can convince the epub wg to adopt it as a reserved prefix (and then get the spec updated, and then get epubcheck updated...).

mattgarrish:
I've moved sourceDate and sourcePublisher to the recommended list. That they don't always apply is a strong case that they shouldn't be required to be present.

mattgarrish:
@clapierre I've tweaked the name of the vocabulary for this document to be more generic. That's probably the best we can do.

(I don't think we want to spilt hairs too finely about what constitutes accessibility metadata as don't forget we only created this space originally because the credential metadata in schema.org wasn't a perfect fit for what we needed.)

PR #275, opened 2024-09-30
Working towards a first version of the Styling Best Practices

bertfrees:
- Move or drop examples
- Drop an (empty) section
- Remove all except one editor's note
- Remove most issues that are not "Seeking an example of ..." issues.
- Improved the "Table of contents with links" example in the Tagging Best Practices (to align with Styling Best Practices)
- Some other small tweaks

See https://github.com/orgs/daisy/projects/8/views/1 for an overview of all the things we may do before we release the first version of the Best Practices (deadline October 11).

***
<a href="https://pr-preview.s3.amazonaws.com/daisy/ebraille/pull/275.html" title="Last updated on Oct 16, 2024, 9:04 PM UTC (e7c2f0a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/daisy/ebraille/275/e3bd853...e7c2f0a.html" title="Last updated on Oct 16, 2024, 9:04 PM UTC (e7c2f0a)">Diff</a>

PR #281, opened 2024-10-01
Add braille examples to most declarations of need in Styling Best Practices

wfree-aph:
Still need to add about five more. Will add those early next week unless I can get to them sooner.

Only added braille. Still need HTML and CSS.


best-practices/styling/index.html
```diff
@@ -885,23 +885,49 @@ <h3>Alignment, positioning and spatial arrangement</h3>
...
+					<pre><code class="formatted-braille">⠠⠽⠕⠥⠀⠺⠊⠇⠇⠀⠗⠑⠚⠕⠊⠉⠑⠀⠞⠕⠀⠓⠑⠁⠗⠀⠞⠓⠁⠞⠀⠝⠕⠀⠀⠀⠀⠀⠼⠉
+⠙⠊⠎⠁⠎⠞⠑⠗⠀⠓⠁⠎⠀⠁⠉⠉⠕⠍⠏⠁⠝⠊⠑⠙⠀⠞⠓⠑⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
+⠉⠕⠍⠍⠑⠝⠉⠑⠍⠑⠝⠞⠀⠕⠋⠀⠁⠝⠀⠑⠝⠞⠑⠗⠏⠗⠊⠎⠑⠀⠺⠓⠊⠉⠓⠀⠀⠀⠼⠙
+⠽⠕⠥⠀⠓⠁⠧⠑⠀⠗⠑⠛⠁⠗⠙⠑⠙⠀⠺⠊⠞⠓⠀⠀⠀⠎⠥⠉⠓⠀⠑⠧⠊⠇⠀⠀⠀⠀⠼⠑
+⠋⠕⠗⠑⠃⠕⠙⠊⠝⠛⠎⠲⠀⠠⠊⠀⠁⠗⠗⠊⠧⠑⠙⠀⠓⠑⠗⠑⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
+⠽⠑⠎⠞⠑⠗⠙⠁⠽⠂⠀⠁⠝⠙⠀⠍⠽⠀⠋⠊⠗⠎⠞⠀⠞⠁⠎⠅⠀⠀⠀⠊⠎⠀⠞⠕⠀⠀⠼⠋
+⠁⠎⠎⠥⠗⠑⠀⠍⠽⠀⠙⠑⠁⠗⠀⠎⠊⠎⠞⠑⠗⠀⠕⠋⠀⠍⠽⠀⠺⠑⠇⠋⠁⠗⠑⠀⠀⠀⠀⠀
+⠁⠝⠙⠀⠊⠝⠉⠗⠑⠁⠎⠊⠝⠛⠀⠉⠕⠝⠋⠊⠙⠑⠝⠉⠑⠀⠀⠀⠊⠝⠀⠞⠓⠑⠀⠀⠀⠀⠼⠛
+⠎⠥⠉⠉⠑⠎⠎⠀⠕⠋⠀⠍⠽⠀⠥⠝⠙⠑⠗⠞⠁⠅⠊⠝⠛⠲⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀</code></pre>
+				</aside>
```
bertfrees, line 925: This example of line numbered text is unfortunately not really an example of inter-word spacing, at least not the way I meant it.

With inter-word spacing I meant when _all_ words in a piece of text are separated by multiple spaces (to increase the readability for beginner braille readers). In this example, you could say that the increased spacing at specific places are "marks", or "gaps" where the numbers would be if they wouldn't be floated to the right.

We have covered this example already in the "[Line numbered prose](https://daisy.github.io/ebraille/best-practices/styling/#example-line-numbered-prose) under the section "Floating content" and I think it kind of fits well with the floating aspect. We could possibly use the example here too, but then I would omit the line numbers (and refer to the other example), and then I would still like an additional example where I can use the [`word-spacing`](https://drafts.csswg.org/css-text/#word-spacing-property) property.

best-practices/styling/index.html
```diff
@@ -1016,17 +1054,180 @@ <h3>Wrapping and breaking text</h3>
...
+⠠⠉⠕⠝⠎⠑⠚⠕⠀⠠⠊⠃⠑⠗⠕⠁⠍⠑⠗⠊⠉⠁⠝⠕⠀⠙⠑⠇⠀⠠⠃⠗⠁⠊⠇⠇⠑⠀
+⠀⠀⠐⠣⠠⠠⠉⠊⠃⠐⠜⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
+⠁⠙⠞⠁⠈⠁⠕⠝⠉⠑⠲⠑⠎⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
+⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠼⠁</code></pre>
+				</aside>
+				<div class="issue" data-number="93" title="Need HTML and CSS of avoiding page breaks within
 				an element"><span/></div>
-
-				<div class="issue" data-number="93" title="Seeking an example of avoiding page breaks between
+				<aside class="example" title="Spatially arranged horizontal equations"></aside>
```
bertfrees, line 1110: > Spatially arranged horizontal equations

Copy-paste error

best-practices/styling/index.html
```diff
@@ -1016,17 +1054,180 @@ <h3>Wrapping and breaking text</h3>
...
+⠓⠊⠎⠀⠺⠗⠊⠞⠊⠝⠛⠲⠀⠠⠺⠓⠑⠞⠓⠑⠗⠀⠊⠞⠀⠺⠁⠎⠀⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
+⠃⠥⠗⠑⠁⠥⠉⠗⠁⠞⠊⠉⠀⠁⠇⠊⠑⠝⠀⠗⠁⠉⠑⠀⠞⠗⠽⠊⠝⠛⠀⠞⠕⠀⠀⠀⠀⠀⠀⠀
+⠙⠑⠍⠕⠇⠊⠎⠓⠀⠠⠑⠁⠗⠞⠓⠀⠞⠕⠀⠍⠁⠅⠑⠀⠺⠁⠽⠀⠋⠕⠗⠀⠁⠝⠀⠀⠀⠀⠀⠀
+⠊⠝⠞⠑⠗⠎⠞⠑⠇⠇⠁⠗⠀⠓⠊⠛⠓⠺⠁⠽⠀⠕⠗⠀⠁⠀⠙⠑⠏⠗⠑⠎⠎⠑⠙⠀⠀⠀⠀⠀</code></pre>
+				</aside>				
+				<div class="issue" data-number="93" title="Need HTML and CSS of avoiding page breaks between
 				two elements"><span/></div>
-
-				<div class="issue" data-number="93" title="Seeking an example of avoiding orphan or widow
+				<aside class="example" title="THE END kept on last line page so it stays with other material (rather than with a blank before)"></aside>
```
bertfrees, line 1158: I'm not sure if we're going to be able to fulfill this requirement with CSS. It can not be done with `page-break-after`.

best-practices/styling/index.html
```diff
@@ -1016,17 +1054,180 @@ <h3>Wrapping and breaking text</h3>
...
+⠠⠉⠪⠄⠎⠀⠠⠍⠊⠇⠅⠒⠀⠠⠮⠀⠍⠕⠌⠀⠺⠊⠙⠑⠇⠽⠀⠥⠎⠫⠀⠍⠊⠇⠅⠀⠀⠀⠀⠀
+⠠⠛⠕⠁⠞⠄⠎⠀⠠⠍⠊⠇⠅⠒⠀⠠⠐⠅⠝⠀⠿⠀⠭⠎⠀⠲⠞⠔⠉⠞⠊⠧⠑⠀⠞⠁⠝⠛⠽⠀
+⠀⠀⠋⠇⠁⠧⠕⠗⠂⠀⠷⠞⠢⠀⠋⠨⠙⠀⠔⠀⠎⠷⠞⠻⠀⠡⠑⠑⠎⠑⠎⠲⠀⠀⠀⠀⠀⠀⠀⠀
+⠠⠩⠑⠑⠏⠄⠎⠀⠠⠍⠊⠇⠅⠒⠀⠠⠗⠊⠡⠀⠔⠀⠋⠁⠞⠂⠀⠹⠀⠍⠊⠇⠅⠀⠀⠀⠀⠀⠀⠀
+⠀⠀⠏⠗⠕⠙⠥⠉⠑⠎⠀⠗⠕⠃⠥⠌⠂⠀⠉⠗⠂⠍⠽⠀⠡⠑⠑⠎⠑⠎⠲⠀⠀⠀⠀⠀⠀⠀⠀⠀
+⠠⠃⠥⠖⠁⠇⠕⠄⠎⠀⠠⠍⠊⠇⠅⠒⠀⠠⠥⠎⠫⠀⠏⠗⠊⠍⠜⠊⠇⠽⠀⠿⠀⠀⠀⠀⠀⠀⠀⠀
+⠀⠀⠍⠕⠵⠵⠜⠑⠇⠇⠁⠂⠀⠃⠥⠖⠁⠇⠕⠄⠎⠀⠍⠊⠇⠅⠀⠁⠙⠙⠎⠀⠁⠀⠗⠊⠡⠂⠀⠀
+⠀⠀⠉⠗⠂⠍⠽⠀⠞⠑⠭⠞⠥⠗⠑⠲⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
+⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠼⠁
+⠀⠀⠀⠀⠠⠡⠑⠑⠎⠑⠀⠠⠍⠁⠅⠬⠀⠠⠏⠗⠕⠉⠑⠎⠎⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
```
bertfrees, line 1215: Would there be enough space for the heading, if it didn't have to be kept with the following text? Doesn't the heading need a blank line before it anyway?

bertfrees:
I have cherry-picked your commit in https://github.com/daisy/ebraille/pull/275.

clapierre (approved):

best-practices/styling/index.html
wfree-aph: Good point. Generating an example that forced this situation was giving me some trouble. It didn't help that I was hand editing the braille some when I noticed that I had left in some print characters that weren't needed. I will update this change in a future commit.

best-practices/styling/index.html
wfree-aph: Thanks @bertfrees . After reviewing `word-spacing`, I'm not sure that it is generally relevant to braille. The closest I can think of is on a device where equidistant pins are utilized, it would be helpful to be told when the spatial arrangement is important so that we can ensure to respect the space between words under all circumstances. With equidistant pins it would make sense to sometimes limit the space between braille cells if it would lead to a large gap (because of the equidistant spacing), like when you have a left-handed cell that is followed by a space and then a right-handed cell. Making the space somewhat smaller does make the space look more like it would on paper but it also affects the layout of the braille so that one line will not line up with the following line like it maybe should. Would this very specific example qualify? If so, something like spatial math would satisfy this requirement.

best-practices/styling/index.html
wfree-aph: @bertfrees Would `page-break-before: avoid` not satisfy the requirement of this kind of formatting?

bertfrees: I don't think so. The margin requirement would have to be violated. I'm going to look for a way to do it, but I don't know if it's possible.

best-practices/styling/index.html
bertfrees: The use case I was thinking of is when it is used in combination with `letter-spacing`, to create blank cells between the braille cells (for unexperienced braille readers). 

When letter-spacing is 1ch, word-spacing could be 2ch or 3ch.

```
W O R D  W O R D
```

or

```
W O R D   W O R D
```

This is a feature that I implemented in Pipeline for Celia (Finland), so it might be relevant for them at least.

Your use case of equidistant pins is also interesting. However, like line numbered prose, I don't know if the `word-spacing` property would be the right way to deal with it. You will need to wrap the two words that should have a reduced word spacing in a span. I guess if an authoring tool could create this markup automatically, it might be fine.

```html
before <span class="reduced-word-spacing">word word</span> after
```

best-practices/styling/index.html
```diff
@@ -1016,17 +1054,180 @@ <h3>Wrapping and breaking text</h3>
...
-				<div class="issue" data-number="93" title="Seeking an example of avoiding page breaks within
+				<div class="issue" title="Need HTML and CSS of forcing page breaks"><span/></div>
+				<aside class="example" title="Forced page break between front matter and body matter"></aside>
+					<p>In order to obtain the following desired layout for the above HTML code,</p>
+					<pre><code class="formatted-braille">⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠁⠰⠊⠧
+⠠⠊⠝⠙⠕⠝⠑⠎⠊⠁⠀⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠀⠼⠛⠚
+⠠⠊⠗⠁⠝⠀⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠀⠼⠛⠁
+⠠⠊⠗⠑⠇⠁⠝⠙⠀⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠀⠼⠛⠉
+⠠⠊⠎⠗⠁⠑⠇⠀⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠀⠼⠛⠙
+⠠⠊⠞⠁⠇⠽⠀⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠐⠀⠼⠛⠋
```
bertfrees, line 1065: Could we change this TOC into something else, e.g. a simple list?

> - Indonesia
> - Iran
> - Ireland
> - Israel
> - Italy

I want as little distraction from what we are trying to show (the `page-break-after` property). (A [table of contents with guide dots](daisy.github.io/ebraille/best-practices/styling/#example-table-of-contents-with-guide-dots) requires a lot of CSS, so a lot of distraction.) Could we also drop the headers/footers with page numbers?

best-practices/styling/index.html
```diff
@@ -1016,17 +1054,180 @@ <h3>Wrapping and breaking text</h3>
...
+⠁⠙⠞⠁⠈⠁⠕⠝⠉⠑⠲⠑⠎⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
+⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠼⠁</code></pre>
+				</aside>
+				<div class="issue" data-number="93" title="Need HTML and CSS of avoiding page breaks within
 				an element"><span/></div>
-
-				<div class="issue" data-number="93" title="Seeking an example of avoiding page breaks between
+				<aside class="example" title="Spatially arranged horizontal equations"></aside>
+					<p>In order to obtain the following desired layout for the above HTML code,</p>
+					<pre><code class="formatted-braille">⠀⠀⠀⠀⠀⠠⠙⠕⠥⠛⠇⠁⠎⠀⠠⠁⠙⠁⠍⠎⠒⠀⠠⠁⠀⠠⠏⠊⠕⠝⠑⠑⠗⠀⠕⠋⠀⠀⠀⠀
```
bertfrees, line 1112: Where should the page breaks go in this example?

best-practices/styling/index.html
```diff
@@ -1016,17 +1054,180 @@ <h3>Wrapping and breaking text</h3>
...
+⠏⠕⠝⠙⠑⠗⠊⠝⠛⠀⠞⠓⠑⠀⠥⠝⠊⠧⠑⠗⠎⠑⠀⠁⠝⠙⠀⠺⠓⠁⠞⠀⠅⠊⠝⠙⠀⠕⠋⠀
+⠛⠥⠊⠙⠑⠀⠎⠕⠍⠑⠕⠝⠑⠀⠺⠕⠥⠇⠙⠀⠝⠑⠑⠙⠀⠞⠕⠀⠝⠁⠧⠊⠛⠁⠞⠑⠀⠊⠞⠲
+⠠⠁⠇⠞⠓⠕⠥⠛⠓⠀⠠⠁⠙⠁⠍⠎⠀⠇⠁⠞⠑⠗⠀⠚⠕⠅⠑⠙⠀⠞⠓⠁⠞⠀⠞⠓⠑⠀⠀⠀
+⠙⠑⠞⠁⠊⠇⠎⠀⠕⠋⠀⠞⠓⠁⠞⠀⠝⠊⠛⠓⠞⠀⠺⠑⠗⠑⠀⠓⠁⠵⠽⠂⠀⠊⠞⠀⠇⠑⠙⠀
+⠞⠕⠀⠞⠓⠑⠀⠉⠗⠑⠁⠞⠊⠕⠝⠀⠕⠋⠀⠞⠓⠑⠀⠃⠑⠇⠕⠧⠑⠙⠀⠋⠊⠉⠞⠊⠕⠝⠁⠇
+⠛⠥⠊⠙⠑⠃⠕⠕⠅⠀⠞⠓⠁⠞⠀⠺⠕⠥⠇⠙⠀⠙⠑⠋⠊⠝⠑⠀⠍⠥⠉⠓⠀⠕⠋⠀⠓⠊⠎⠀
+⠉⠁⠗⠑⠑⠗⠲⠀⠠⠞⠓⠊⠎⠀⠍⠕⠍⠑⠝⠞⠀⠕⠋⠀⠎⠑⠗⠑⠝⠙⠊⠏⠊⠞⠽⠀⠊⠎⠀⠀
+⠕⠋⠞⠑⠝⠀⠉⠊⠞⠑⠙⠀⠁⠎⠀⠕⠝⠑⠀⠕⠋⠀⠞⠓⠑⠀⠍⠕⠎⠞⠀⠋⠁⠍⠕⠥⠎⠀⠀⠀
+⠦⠑⠥⠗⠑⠅⠁⠴⠀⠍⠕⠍⠑⠝⠞⠎⠀⠊⠝⠀⠇⠊⠞⠑⠗⠁⠗⠽⠀⠓⠊⠎⠞⠕⠗⠽⠲⠀⠀⠀
+⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛⠛
```
bertfrees, line 1147: Should I use a  `<div class="box">` for this?

bertfrees:
@wfree-aph Do you have time to work on this today? If possible I would like to do most of what remains to do for Best Practices 1.0 today.

best-practices/styling/index.html
bertfrees: @wfree-aph I've been looking at the rest of the example, trying to figure out the HTML, but I need some help with it.

Or perhaps we just need a simplified example, it seems that for this example there will be quite a lot of different markup and styling. We could possibly just drop everything after the title "World Braille Usage Third Edition Braille Usage by Country".

(By the way, I would even drop the "(cont.)", as this is something related to volume breaking, something we haven't touched yet, as we're focusing on electronic reading systems now, and will explore embossing only later).

best-practices/styling/index.html
```diff
@@ -1016,17 +1054,180 @@ <h3>Wrapping and breaking text</h3>
...
+⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠨⠣⠠⠞⠓⠑⠀⠠⠑⠝⠙⠈⠨⠜⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀</code></pre>
+				</aside>
+				<div class="issue" data-number="93" title="Need HTML and CSS of avoiding orphan or widow
 				lines"><span/></div>
-
+				<aside class="example" title="Heading 2 moving to next page to remain with material"></aside>
+					<p>In order to obtain the following desired layout for the above HTML code,</p>
+					<pre><code class="formatted-braille">⠠⠔⠞⠗⠕⠙⠥⠉⠰⠝⠀⠞⠕⠀⠠⠡⠑⠑⠎⠑
+⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
+⠀⠀⠠⠡⠑⠑⠎⠑⠀⠊⠎⠀⠐⠕⠀⠷⠀⠮⠀⠕⠇⠙⠑⠌⠀⠯⠀⠍⠕⠌⠀⠧⠻⠎⠁⠞⠊⠇⠑⠀
```
bertfrees, line 1192: Like in the "Forced page break between front matter and body matter" example, there are a lot of things going on in this example as well: different types of headings, paragraphs and list items. Would benefit from some simplification (to better focus on what we're trying to show).

best-practices/styling/index.html
bertfrees: I still haven't found a solution yet. The [fragmentation module](https://www.w3.org/TR/css-break-3) doesn't seem to account for this use case. For regular (screen) CSS, one might be able to achieve something with [flexbox](https://www.w3.org/TR/css-flexbox), but for braille, there is the additional complexity that margins need to be multiples of braille cells, so I don't know if I should even attempt to explore this path.

bertfrees: Going to comment out the example for now.

best-practices/styling/index.html
bertfrees: I've changed it to "Box kept together on page".

best-practices/styling/index.html
bertfrees: Using a `<div class="box">` in 299f6a4.

best-practices/styling/index.html
bertfrees: The example is commented out for now.

best-practices/styling/index.html
wfree-aph: @bertfrees Sorry for my delay. A table of contents with guide dots is the requirements of a region of braille formatting. I don't know that we can just drop it entirely. You could leave it as a noted limitation and try to figure it out later though.

Some regions will make a simple list, so that will make sense. It will show how to make the list and include the links to those page numbers.

best-practices/styling/index.html
wfree-aph: I think I understand the question but please correct me if I'm wrong. You're avoiding page breaks within the box. So including and between the row of dots 2356 and the row of dots 1234.

best-practices/styling/index.html
wfree-aph: Would you feel comfortable just deleting any aspects that seem unnecessary or are you asking me to go in and edit it? The only things you need are the last line before the end of the braille page, the heading itself, and then the line after the heading itself. The braille page break could be indicated in some other way but in the current example, it is done using a braille page number. If you have another suggestion here, I think that would work.

best-practices/styling/index.html
bertfrees: The point is that this example is not about a TOC, it's about making the bodymatter start on a new page.

I don't want to mess with the braille formatting requirements, I was thinking we would just change the markup so that it wouldn't be a TOC anymore, but something else with simpler CSS.

The other solution is to omit any CSS that is not relevant to the feature that the example is actually about. It would be a departure from how it was done until now in the rest of the document, but it could work.

wfree-aph: Ah. My fault for misunderstanding. In that case, making it a simple list works entirely. Just remove the guide dots and page numbers. At that point it's just a list of country names but takes up the same amount of space.

best-practices/styling/index.html
bertfrees: Yes, sorry, quite obvious.

best-practices/styling/index.html
bertfrees: No, not asking you to go in and actually edit it. Just wanted to know what you think before I would go mess with the example.

Yes, I think it would be good to simplify/condense the example a bit. I think I can figure it out.

best-practices/styling/index.html
bertfrees: @wfree Can you help me with the markup of the first elements in the bodymatter? No need to give me HTML code. Just tell me what these element are:

1. ```
          ,WORLD ,BRAILLE ,USAGE
               ,THIRD ,EDITION           
   ```
2. ```
   ,BRAILLE ,USAGE BY ,COUNTRY "<CONT4"> 
   ```
3. ```                                      
                   ,ECUADOR              
   ```
4. ```
   ,LANGUAGE3 ,SPANISH "<USUALLY UNCONTRACTED">                      
   ```
5. ```
    ,BRAILLE ,STANDARDS ,SET BY3
   ```      
6. (I assume this is a list)
   ```
   ,LATIN ,AMERICAN ,BRAILLE ,COUNCIL    
     "<,COMISI^/ON ,BRAILLE              
     ,LATINOAMERICANA.> "<,,CBL">        
   ,LATIN ,AMERICAN ,BLIND ,UNION        
     "<,UNI^/ON ,LATINOAMERICANA DE      
     CIEGOS"> "<,,ULAC">                 
   WWW4ULACDIGITAL4ORG                   
   ,BRAILLE@AULACDIGITAL4ORG             
   ```

best-practices/styling/index.html
wfree-aph: No worries! I could have been more clear with a comment or something.

best-practices/styling/index.html
wfree-aph: 1 through 3 are all separate h1s (centered headings). 4 is a solitary list item. 5 is an h2 (cell-5 heading). And 6 is a list as you assumed.

best-practices/styling/index.html
bertfrees: I ended up just adding an extra empty line. So now there are 26 lines on the page (but no one will notice).

best-practices/styling/index.html
bertfrees: By the way another issue with example is that it is in contracted braille. I think all other examples are uncontracted (and it's also what we requested). But we can fix that another time.

best-practices/styling/index.html
bertfrees: Thanks. I ended up keeping the TOC and most of the content on the next page. In the CSS I added this comment:

```css
/* For clarity, any CSS that is not relevant for the page break was left out */
```

bertfrees:
Going to close this PR now as it was incorporated in https://github.com/daisy/ebraille/pull/275.

PR #282, opened 2024-10-02
Replace custom source metadata with dublin core

### LINKED ISSUE #280: Use refines for source metadata
When I look at having to define two new properties for the source publication, I wonder if we're following the flat metadata model from epub 2 / daisy 3 too much.

If we had a structured metadata model, all the source information would ideally be kept together. It might be structured something like this:

```
"source": {
   "name": "Moby Dick",
   "identifier": "urn:isbn:###",
   "publisher": "Acme Publishing",
   "date": "2000-01-01"
   ...
}
```

Although EPUB 3's refines attribute isn't the greatest thing ever invented, it is a means of connecting related metadata statements and in our case would let us re-use dublin core metadata rather than invent new properties. For example, instead of dc:source, a11y:sourcePublisher, and a11y:sourceDate as three unconnected properties, we could attach existing dcterms properties:

```
<dc:source id="src">urn:isbn:###</dc:source>
<meta property="dcterms:publisher" refines="#src">Acme Publishing</meta>
<meta property="dcterms:date" refines="#src">2001-01-01</meta>
...
```

In the definition for dc:source we could recommend these properties be used for the publisher and date so that there is guidance to publishers on how to handle this with the refines mechanism.

mattgarrish:
This pull request removes sourcePublisher and sourceDate from the vocabulary and replaces them with dcterms:publisher and dcterms:date. Their usage is integrated into the section on dc:source so the information is all in one place.

Fixes #280

* [Preview](https://raw.githack.com/daisy/ebraille/spec/issue-280/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/issue-280/index.html)


wfree-aph (approved):
Thanks for thinking of this change. It seems cleaner and more appropriate.

bertfrees (approved):

PR #283, opened 2024-10-07
Add recommendations against using css that adds non-braille characters

### LINKED ISSUE #273: Styling best practices: drop the section on "automatic numbering" for now
Since we are probably going to recommend hard-coding the numbers/bullets of list items and headings into the text content, we might as well drop the section on "Automatic numbering" for now. There is nothing else to be numbered: we haven't covered braille page numbers and footnotes in the best practices yet (deliberately omitted for now because of bad/lacking support in browsers).

By the way, we might need an example of `list-style-type: none` in the best practices.

mattgarrish:
This pull request adds a recommendation to the xhtml character encoding section not to use CSS that would add non-braille characters to the text.

It also adds a section on formatting xhtml ordered and unordered lists with a recommendation to embed the list numbering rather than rely on the default styling.

There's also a minor fix to a broken link and I changed the word "required" to "expected" at the start of the section on character encoding so it doesn't sound contradictory to the actual normative recommendation we have about the use of braille characters.

Related to #273

### DIFF
--- a/index.html
+++ b/index.html
@@ -1820,7 +1820,7 @@
 
 					<p>The restriction also applies to CSS styling that would generate non-braille characters (e.g.,
 						list styling for numbers, letters, and glyphs, and the use of the <code>content</code> property
-						to dynmically add text before or after elements). Reading systems are not expected to convert
+						to dynamically add text before or after elements). Reading systems are not expected to convert
 						such styling to braille characters.</p>
 
 					<div class="note">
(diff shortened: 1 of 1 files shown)


mattgarrish:
* [Preview](https://raw.githack.com/daisy/ebraille/spec/css-numbering/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/css-numbering/index.html)

clapierre (approved):

index.html
```diff
@@ -1818,7 +1818,16 @@ <h4>Character encoding</h4>
 						<li>the [^abbr/title^] attribute [[html]] for expansions of abbreviations.</li>
 					</ul>
 
-					<p>This restriction does not apply to attributes that carry meta information (e.g., the
+					<p>The restriction also applies to CSS styling that would generate non-braille characters (e.g.,
+						list styling for numbers, letters, and glyphs, and the use of the <code>content</code> property
+						to dynmically add text before or after elements). Reading systems are not expected to convert
```
bertfrees, line 1823: typo: must be "dynamically"

index.html
```diff
@@ -1828,6 +1837,18 @@ <h4>Character encoding</h4>
...
+				<section id="xhtml-list-format">
+					<h4>List formatting</h4>
+
+					<p>When authoring [[html]] ordered ([^ol^]) and unordered ([^ul^]) lists, eBraille creators SHOULD
+						embed the list number, letter, or glyph within each item (e.g., by setting
+							<code>list-style-type</code> to <code>none</code> to override the default styling). Do not
+						rely on [=reading systems=] to convert the automatic styling of lists to braille characters.</p>
+
+					<p>CSS SHOULD only be used for styling lists when the braille characters to display are supplied
+						(e.g., by setting <code>list-style-type: '⠿⠲ '</code> to insert braille bullets).</p>
```
bertfrees, line 1849: > CSS SHOULD only be used for styling lists when the braille characters to display are supplied.

I had two minor issues reading this sentence.

- > be used for styling lists

  To me "list styling" is a bit more general than what we refer to here (even though CSS has the `list-style` property for exactly this). Styling could also include spacing and indentation, but we only mean setting markers. The best alternative I can come up with is "marking of list items". "List numbering" would be too specific (as it does not include bullets).

- > when the braille characters to display are supplied

  At first I didn't understand this sentence. Supplied by whom? ... Oh I see, supplied by the CSS style sheet. It could be just me though. For me this would be clearer:

  > .when the style sheet specifies the braille characters to display.

bertfrees (approved):

wfree-aph (approved):

PR #302, opened 2024-12-27
Styling best practices for tables

bertfrees:
I've added all the collected examples for tables, except for the the one for the "listed" BANA table format, because it can not be handled with CSS alone (due to the table headers being inserted before each cell).

The examples are divided over the sections "Alignment, positioning and spatial arrangement", "Borders and lines" and "Alternate table formats".

[Preview](https://raw.githack.com/daisy/ebraille/styling-tables/best-practices/styling/index.html?developer-mode=true)


wfree-aph (approved):
I do not see any issues. Thanks for thinking of these solutions. I know it's complicated. Apologies for the delay in reviewing. It came through while I was on vacation and I just missed it after that.

PR #303, opened 2025-02-11
Remove editor notes

mattgarrish:
This pull request changes the "should not" to a "must not" for authoring media queries using the braille media type. I also added an explicit restriction against using the grid feature as it was only mentioned in the following note.

As a consequence, I've removed the section on reading system handling of these values, as if they're not allowed for authoring then we shouldn't be recommending reading systems accommodate them.

But let me know if this makes sense to you, @bertfrees 

I also removed the two other editor notes about possible future directions, as these aren't necessary in the document. If there are problems with the object element or we need to say more about landmarks, we should get proper issues to address them.

And finally, I removed the aliasing of the version-less CSS modules. Respec was changed a while back to automatically match these to the latest versions.

***

[Preview](https://raw.githack.com/daisy/ebraille/fix/ed-notes/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/fix/ed-notes/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -74,30 +74,6 @@
 						"title": "Digital Publishing WAI-ARIA Module",
 						"href": "https://www.w3.org/TR/dpub-aria/",
 						"publisher": "W3C"
-					},
-					"css-page": {
-						"aliasOf": "css-page-3"
-					},
-					"css-fonts": {
-						"aliasOf": "css-fonts-3"
-					},
-					"css-color": {
-						"aliasOf": "css-color-3"
-					},
-					"css-text-decor": {
-						"aliasOf": "css-text-decor-3"
-					},
-					"css-values": {
-						"aliasOf": "css-values-3"
-					},
-					"mediaqueries": {
-						"aliasOf": "mediaqueries-4"
-					},
-					"css-conditional": {
-						"aliasOf": "css-conditional-3"
-					},
-					"css-cascade": {
-						"aliasOf": "css-cascade-5"
 					}
 				}
 			};
@@ -424,10 +400,6 @@
      aria-label="alternative text">
    &lt;p>&lt;!-- description -->&lt;/p>
 &lt;/object></pre>
-				</aside>
-				<aside class="ednote">
-					<p>The accessibility of using an <code>object</code> element is still to be determined. Guidance on
-						how to embed PDFs may change in a future update.</p>
 				</aside>
 			</section>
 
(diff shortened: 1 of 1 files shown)


bertfrees:
Forbidding the `braille` media type makes total sense, we don't want to recommend using deprecated features:

> Media queries MUST NOT include the deprecated <code>braille</code> media type to target braille displays.

However I don't think the other changes are an improvement. This sentence that was dropped, captured exactly what we wanted to achieve, I think:

> eBraille creators SHOULD NOT use media queries that discriminate between a braille display and a screen when the target medium is a braille display.

(I'm not saying the wording could not be improved.)

By including this requirement we wanted to ensure that:

1. how an eBraille publication shows on a computer screen is representative for how it will show on a braille display.
2. we are not giving implementors of eBraille reading systems, which are likely going to be based on existing CSS implementations, a hard time.

We must keep these goals in mind and add only the necessary restrictions. By replacing the sentence with the `grid` restriction,

- the intent becomes less clear IMO
  - you need to read the note to understand
  - too much importance is given to `grid`, while there might be other media features that discriminate between a screen and a braille display (notably the `screen` type, which we recommend against in the note)
- I think we add too big of a restiction
  - something that was recommended against is now forbidden
  - it is not clear enough whether `grid` is forbidden altogether, or whether it is allowed in certain cases, e.g. to target an embosser
  - it goes against our wish/goal that in the future user agents can actually support differences in styling for screens vs. braille displays

The latter is an important one for me.

I agree of course that it would be a nice simplification if we could drop the requirements on the reading system side, and your change does that. It would be good if we could drop this slightly cryptic sentence, but I don't know if that's possible at the moment:

> eBraille reading systems of type braille display SHOULD match both <code>(grid)</code> and <code>not (grid)</code>

mattgarrish:
I've made some more fixes to try and handle your comments. I find the long note in that section confusing because it introduces recommendations that aren't mentioned in the text. That's why I added the "must not" for grid, but there's also a recommendation against using screen buried in it, too.

I've instead turned that section into a bullet list with all the requirements. I kept the two additional explanatory paragraphs from the note as regular text below the list. I've also restored the reading system section with the grid support recommendation.

Let me know if this works better, or feel free to edit the text yourself, as you're in a better position to know what should be said here.

bertfrees:
Thanks, that's better!

I'm going to edit the text some more. If it's OK with you, I'd like to:

- make it clear that certain requirements only apply for media queries that target a braille display
- integrate the text below the bullet list into the list, to make it more clear which motivations and remarks belong to which requirements
- make it more clear that "to be forward-compatible with a possible future version ..." is the reason for discouraging the use of `screen` for targetting a braille display (not a reason for using `screen`)

Is the `braille` media type "deprecated", or "expected to be deprecated"? "Expected to be deprecated" is what's in the last sentence, and is how I initially put it, because Media Queries Level 4 is a draft. "Deprecated" is what's in the first bullet point.

mattgarrish:
> I'm going to edit the text some more. If it's OK with you

Yes, I was just hoping for an easy removal of the editor notes. Please fix the text up however you think it should best be stated.

bertfrees:
@mattgarrish I pushed my attempt to further improve the media query sections.

PR #306, opened 2025-02-25
Various fixes from feedback

mattgarrish:
This pull request addresses the feedback received on list from @jylioja21

Most notably:

- it removes the `dateTranscribed` property
- it rewords the note about how an ebraille file with a .epub extension would be handled. This is tricky to say definitively as the media type could be the determining factor if the file is served over the web, for example. And unless the user is already in a reading system, it might be the operating system that routes the file to the wrong type of reading agent based on its extension.

***

[Preview](https://raw.githack.com/daisy/ebraille/spec/typos/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/typos/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1675,7 +1675,7 @@
    &lt;manifest>
       &lt;item id="c01"
             href="eBraille/chapter01.xhtml"
-            media-type="application/xhtml+xml"
+            media-type="application/xhtml+xml"/>
       &#8230;
    &lt;/manifest>
    &lt;spine>
--- a/index.html
+++ b/index.html
@@ -2518,8 +2518,8 @@
 			<div class="note">
 				<p>If an eBraille file is packaged with the extension <code>.epub</code>, it might not be recognized as
 					an eBraille publication (e.g., if it is also not served with the <a href="#app-media-type">eBraille
-						media type</a> or that media type is not recognized). In this case, the file would likely be
-					opened as an EPUB publication in an EPUB reading system with some loss of braille-specific
+						media type</a> or that media type is not recognized). In this case, the file would likely only
+					open as an EPUB publication in an EPUB reading system, resulting in some loss of braille-specific
 					functionality.</p>
 			</div>
 
--- a/index.html
+++ b/index.html
@@ -1081,7 +1081,7 @@
 
 						<aside class="example" title="Copyright date where only the year is known">
 							<pre>&lt;meta property="dcterms:dateCopyrighted">
-   2014
+   2004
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1539,8 +1511,9 @@ <h4>Examples</h4>
          41f1328c-0571-4e71-8be8-e65bc148281a
       &lt;/dc:identifier>
       &lt;dc:language>en-Brai-US&lt;/dc:language>
+      &lt;dc:date>2014-12-11T09:12:52+00:00&lt;/dc:date>
       &lt;meta property="dcterms:dateCopyrighted">
-         2024
+         2004
```
clapierre, line 1516: why did this change from 2024 to 2004?

mattgarrish: The copyright date is of the original work. World Cultures and Geography was published in 2004, so that's the correct year.

The dc:date of 2014 is when the epub 3 sample this metadata came from was created. I think we can update that to 2025 since we're mocking up data for an ebraille version of it.

mattgarrish: Should dateCopyrighted be recommended and not required?

What if you're creating an original braille work? What would you set this field to in that case?

jylioja21 (approved):
Original sentence: “If an eBraille file is packaged with the extension .epub, it will not be recognized by eBraille reading system as an eBraille publication.”
Matt commented: "it rewords the note about how an ebraille file with a .epub extension would be handled. This is tricky to say definitively as the media type could be
the determining factor if the file is served over the web, for example. And unless the user is already in a reading system, it might be the operating system
that routes the file to the wrong type of reading agent based on its extension."
My main concern with how it was previously written was changing the singular "reading system" to plural "reading systems". But if it's clearer with the additional explanation, I'm ok with that too. Whichever way you prefer.

mattgarrish:
> My main concern with how it was previously written was changing the singular "reading system" to plural "reading systems".

Right, when I went to make your change I ended up not liking what the note was saying. It made the situation of an ebraille file without the required extension sound too simplistic. The extension is one means of determining what kind of file it is, but there's also the media type if the file is being served over the web. The operating system and/or reading system could still handle the file correctly if the extension is ignored in favour of the media type, which is often how it works. Hopefully, the new wording captures this a bit better.

PR #315, opened 2025-04-22
Create folder for exemplar starting files and metadata.txt

wfree-aph:
Adds folder for starting files for exemplars and TXT file with metadata requirements needed for each starting file

### DIFF
--- a/experiments/exemplars/starting files/Metadata.txt
+++ b/experiments/exemplars/starting files/Metadata.txt
@@ -1,0 +1,26 @@
+Please include the following with your submission:
+
+Required-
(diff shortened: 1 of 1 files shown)


skntkacm:
[slovak_examples.zip](https://github.com/user-attachments/files/19888921/slovak_examples.zip)
Here are the examples from Slovakia. Examples are zipped along with a file with metadata and description. 
@wfree-aph  Please, let me know, if this was delivered correctly and can be used. If there is something I can do, please, send me an e-mail.

wfree-aph:
Went ahead and merged this since it was so basic.

PR #327, opened 2025-05-06
clarify the relationship to EPUB

### LINKED ISSUE #316: Clarify relationship to EPUB?
I find the eBraille specification is not always very clear about how it —normatively— to EPUB.

## How eBraille/EPUB is currently defined

As a reminder, the way it currently articulates is as follows:

- the normative relationship to EPUB stems from the "eBraille publication conformance" section, which —similar to how it's done in EPUB— contains normative statements on the high-level MUSTs pointing to subsequent eBraille sections (package document, primary entry page, publication resources, packaging, etc).
- then, each of the eBraille main sections typically contain a normative statement linking to EPUB. For instance for the package document:

> At the root of the eBraille package document is the package element. This element MUST conform to the [requirements for the package element](https://www.w3.org/TR/epub/#sec-package-elem) in [[epub3](https://daisy.github.io/ebraille/#bib-epub3)].

**Note**: eBraille better explains the relationship to EPUB in the "Relationship to EPUB 3" section, but this is non-normative

## Some issues

1. some normative statements in EPUB are not propagated to eBraille

For instance, for package documents, by only normatively referencing to the requirements for the package element in EPUB 3, eBraille in some way leaves out EPUB's normative "Parsing URLs in the package document" section.

2. some dependencies to EPUB are not expressed as RFC2119 statements

For instance, in the "XML conformance" section:

> The requirements for XML-based media types [[rfc2046](https://daisy.github.io/ebraille/#bib-rfc2046)] defined in [XML conformance](https://www.w3.org/TR/epub/#sec-xml-constraints) [[epub3](https://daisy.github.io/ebraille/#bib-epub3)] also apply to [eBraille publications](https://daisy.github.io/ebraille/#dfn-ebraille-publication).

or in the "Resource location" section:

> eBraille does not support the file: URL scheme [[rfc8089](https://daisy.github.io/ebraille/#bib-rfc8089)] for referencing resources in an eBraille publication.

## Suggestion for improvement?

I'm wondering if the entire relationship to EPUB could be made more explicit if eBraille said something like:

> a conforming eBraille publication MUST be a conforming EPUB publication, except for the [willful violations](https://infra.spec.whatwg.org/#willful-violation) documented in this specification.

using the [willful violation](https://infra.spec.whatwg.org/#willful-violation) terminology form Infra.

The only willful violation, as far as I can tell, being that eBraille does not require to be packaged as an OCF ZIP Container.

All this is mostly editorial, but I believe it would clarify some normative bits. Also, it may be easier to maintain since anything added to EPUB, any editorial change in EPUB's spec organization, will be automatically ported to eBraille.

@mattgarrish what do you think?

### LINKED ISSUE #318: Editorial: remote publication resources
The statement in "4.2 File and directory structure":

> Publication resources MUST NOT be located outside the publication root or its descendant directories.

is duplicated in "4.4 URLs in the file set"

> eBraille file set MUST NOT reference publication resources outside the [publication root](https://daisy.github.io/ebraille/#fileset-structure).

Also, I suggest replacing "as defined" by "as described" in the following 4.4 sentence:

> [Remote resources](https://www.w3.org/TR/epub/#dfn-remote-resource) [[epub3](https://daisy.github.io/ebraille/#bib-epub3)] are not supported, as defined in [3.5 Resource location](https://daisy.github.io/ebraille/#res-location).

since the normative statement is in "4.2 File and directory structure" and not "3.5 Resource location".

Note that these specific editorial suggestions may become moot depending on the course of action taken for #316.

rdeltour:
This PR attempts to address #316 by making the following changes:

- **publication conformance:**
  - add a conformance statement saying eBraille MUST conform to EPUB except for documented willful violations
  - change the way we refer to packaging requirements (see below)
- **eBraille file set:**
  - make eBraille file set a virtual file set that MUST conform to OCF abstract container
  - define two materializations of the eBraille file set in two new subsections, **Packaged eBraille file set** and **Unpackaged eBraille file set**.
  - move the publication resources location conformance statement to the “Publication resources” section. The “URLs in the file set” section becomes informative (it inherits from the EPUB one).
- **deployment:**
  - replace the “Web deployment” section by a more generic “Deployment” with two sub-section, “Packaged eBraille publications” and “Web- hosted eBraille publications”.
  - this section therefore contains all the conformance statements related to actual distribution, which are not generically testable.

That’s the general idea. It basically started as an experiment but eventually turned out to be a fairly serious edit. I find it adds some clarity, especially as it sticks closer to EPUB, but any thoughts welcome!

To be refined if we decide to move forward with the PR:
- maybe we need new terms in the terminology for packaged/unpackaged file sets
- check the markup (citations, definitions, etc), as I’m not well versed in ReSpec.
- formatting (I’m not sure what tool you usually use for formatting?)

Closes #316 
Closes #318

* * *

[Preview](https://raw.githack.com/daisy/ebraille/spec/issue-316/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://daisy.github.io/ebraille/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/daisy/ebraille/spec/issue-316/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -351,7 +351,8 @@
 						<p>This is a <a data-cite="infra#willful-violation">willful violation</a> of
 							<a data-cite="epub3/#sec-container-file-and-dir-structure">EPUB publication conformance</a>
 							[[epub3]], which requires a publication to be packaged in an <a 
-							data-cite="epub3#dfn-epub-container">EPUB container</a>.</p>
+							data-cite="epub3#dfn-epub-container">EPUB container</a>, to facilitate the deployment of
+							eBraille publications from the web without users having to download a ZIP file.</p>
 					</div></li>
 			</ul>
 
--- a/index.html
+++ b/index.html
@@ -546,7 +546,7 @@
 					have a true file system within it, the rules for file naming and placement can only be defined
 					virtually.</p>
 
-				<p>Unlike EPUB, eBraille allows materialization of the OCF abstract container both as a physical file set
+				<p>Unlike EPUB, eBraille allows manifestation of the OCF abstract container both as a physical file set
 					and as a packaged container. This allows an [=eBraille publication=] to be independent
 					of packaging, making it a format that can flexibly move between web deployment and end-user
 					distribution.</p>
@@ -652,7 +652,7 @@
 			<section id="ebrl-unpackaged-fileset">
 				<h3>Unpackaged eBraille file set</h3>
 
-				<p>An <dfn>unpackaged eBraille file set</dfn> is a multi-file materialization of the
+				<p>An <dfn>unpackaged eBraille file set</dfn> is a multi-file manifestation of the
 					[=eBraille file set=] where:</p>
 
 				<ul>
@@ -680,7 +680,7 @@
 			<section id="ebrl-packaged-fileset">
 				<h3>Packaged eBraille file set</h3>
 
-				<p>A <dfn>packaged eBraille file set</dfn> is a single-file materialization of the [=eBraille file set=] that:<p>
+				<p>A <dfn>packaged eBraille file set</dfn> is a single-file manifestation of the [=eBraille file set=] that:<p>
 
 				<ul>
 					<li>MUST be a conforming <a data-cite="epub3/#sec-container-zip">OCF ZIP container</a>.</li>
--- a/index.html
+++ b/index.html
@@ -333,6 +333,9 @@
 			<p>A conforming [=eBraille publication=]:</p>
 
 			<ul>
+				<li>MUST be a conforming <a data-cite="epub3#dfn-epub-publication">EPUB publication</a> [[epub3]],
+					except for the <a data-cite="infra#willful-violation">willful violations</a> [[infra]]
+					documented in this specification.</li>
 				<li>
 					<p>MUST define at least one rendering of its content as follows:</p>
 					<ul>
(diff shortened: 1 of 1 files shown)


rdeltour:
(edited to add preview and diff links)

clapierre (approved):

index.html
```diff
@@ -526,16 +538,16 @@ <h2>eBraille file set</h2>
...
 						data-cite="epub3#sec-container-abstract">OCF abstract container</a> [[epub3]]. EPUB 3 only
 					defines its file set in the abstract because those files are expected to be zipped in the <a
 						data-cite="epub3#dfn-epub-container">EPUB container</a> [[epub3]]; the standard is not concerned
 					with the physical files before they are zipped or after they are unzipped. As a ZIP file does not
 					have a true file system within it, the rules for file naming and placement can only be defined
 					virtually.</p>
 
-				<p>eBraille moves the rules for the OCF abstract container to the physical file set that exists before
-					and after packaging in the EPUB container. This allows an [=eBraille publication=] to be independent
+				<p>Unlike EPUB, eBraille allows materialization of the OCF abstract container both as a physical file set
```
mattgarrish, line 549: Minor nit, but "materialization" has a sudden appearance ring to it (often used for ghosts appearing). Maybe these uses should be "manifestation" to match the EPUB spec.

index.html
```diff
@@ -342,14 +345,19 @@ <h2>eBraille publication conformance</h2>
...
 				</li>
 				<li>SHOULD conform to the accessibility requirements defined in <a href="#ebrl-a11y"></a>.</li>
+				<li>MUST be either an [=unpackaged eBraille file set=] or a [=packaged eBraille file set=] as defined
+					in <a href="#ebrl-fileset"></a>.
+					<div class="note">
+						<p>This is a <a data-cite="infra#willful-violation">willful violation</a> of
+							<a data-cite="epub3/#sec-container-file-and-dir-structure">EPUB publication conformance</a>
+							[[epub3]], which requires a publication to be packaged in an <a 
+							data-cite="epub3#dfn-epub-container">EPUB container</a>.</p>
+					</div></li>
```
mattgarrish, line 355: Another minor nit, but infra says to note the reason for any willful violations so I wonder if we should say that this is to allow publishers to deploy ebraille publications from the web? I'm fine either way, though, as I'm sure we must make this case elsewhere.

mattgarrish (approved):
Thanks, these changes are great! I've just added a couple of very minor suggestions..

I didn't spot that this adds any new requirements, but we should probably still document the change given the scope. I can add a change log entry after merging if you want.

index.html
rdeltour: sure, "manifestation" sounds good!

index.html
rdeltour: Right. I didn't add the motivation because it was explained in informative text elsewhere, but it's distant (in the file set intro) so it's better to make it explicit in this note!

rdeltour:
I just pushed edits following the suggestions you made 👍

> I didn't spot that this adds any new requirements, but we should probably still document the change given the scope. I can add a change log entry after merging if you want.

Right, this shouldn't add any new requirements (none that weren't already in EPUB), but a change log entry sounds good to me!

mattgarrish:
Thanks again for this! I'll leave it open until end of day tomorrow just to give some more time for review, but I'm not too concerned about merging since we're only clarifying what we had always intended.

rdeltour:
No problem! Sounds good.

